PR #2, opened 2019-05-17
feat(dropdown): add a11y interactions

cal-smith:
Addressed all the outstanding comments

Reopening since I mistyped and pushed to the wrong branch.


src/components/dropdown/dropdown.js
```diff
@@ -178,6 +228,9 @@ class BXDropdown extends HTMLElement {
   connectedCallback() {
     this.attachShadow({ mode: 'open' });
     this.render();
+
+    // listen for clicks on the body so we can close the dropdown
+    this.ownerDocument.addEventListener('click', this.#handleFocusIn);
```
asudoh: Seems that my earlier suggestion on bubbling (capturing) `focus` event has to be replaced with `blur`, my apologies for the confusion. Below code on top of yours seems to work instead. Would you try...?

```javascript
#handleFocusOut = event => {
  if (!this.contains(event.relatedTarget)) {
    this.open = false;
  }
};

connectedCallback() {
  ...

  // Detect IE/Edge which have bubbling `window.onfocusout`
  const hasFocusOut = 'onfocusout' in this.ownerDocument.defaultView;
  const focusoutEventName = hasFocusOut ? 'focusout' : 'blur';
  // Use `focusout` if it's there, otherwise use "capture" mode which has similar-to-bubbling effect
  this.ownerDocument.body.addEventListener(focusoutEventName, this.#handleFocusOut, !hasFocusOut);
}
```

src/components/dropdown/dropdown-item.js
```diff
@@ -82,13 +119,27 @@ class BXDropdownItem extends HTMLElement {
...
+  }
+
   /**
    * The CSS class for the selected item.
    * @type {string}
    */
   static classSelected = `${prefix}--dropdown--selected`;
 }
 
+focusMixin(BXDropdownItem);
```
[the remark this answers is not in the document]
cal-smith: whoops! yep that's unintentional

src/components/dropdown/dropdown.js
cal-smith: I'll try that out ... I'm also not sure what happened with that change 😱 definitely not what I intended

cal-smith: For `Escape` I still had to handle the `.open = false` in `#handleKeyDownInner` - testing in storybook it doesn't return focus to the dropdown button (I think it puts focus out onto the iframe?), but testing in isolation it behaves as it should.

cal-smith: Tab also has some weird behaviour ... `Tab` moves the focus fine, but `shift + Tab` seems to move the focus to the trigger (without activating it's focus state) and then a second `shift + Tab` moves it out of the component

cal-smith:
Updated ... there's a few weird things now with focus handling

asudoh:
Thank you for the update @cal-smith! Escape key is a good point - I found that Storybook seems to take over the focus upon Escape key which caused my false assumption that Escape key causing `blur` (🙇). Another possible cause of the weird focus behavior you saw is that `.relatedTarget` contains no hint on whether the focus went to the trigger button or not, because the trigger button is in shadow DOM and the Shadow DOM spec avoids leaking out the shadow DOM details in `.relatedTarget`.

BTW if it makes sense to you, I'd like `#toggle()` used for user-initiated open/close, whereas `.open = newOpenState` used for system-initiated open/close, because user-initiated open/close associates with change in focus.

Below is my suggestion reflecting above, great if you can try - Thanks!

```diff
diff --git a/.storybook/config.js b/.storybook/config.js
index 2e44b1c..eac72fd 100644
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -22,4 +22,11 @@ function loadStories() {
   req.keys().forEach(filename => req(filename));
 }
 
+document.addEventListener('keydown', event => {
+  if (event.key === 'Escape') {
+    // Prevent presumable Storybook behavior taking over focus upon `Escape`` key
+    event.stopPropagation();
+  }
+});
+
 configure(loadStories, module);
diff --git a/src/components/dropdown/dropdown.js b/src/components/dropdown/dropdown.js
index d549eff..9299b43 100644
--- a/src/components/dropdown/dropdown.js
+++ b/src/components/dropdown/dropdown.js
@@ -65,8 +65,11 @@ class BXDropdown extends HTMLElement {
   };
 
   #handleFocusOut = event => {
-    if (!this.contains(event.relatedTarget)) {
-      this.#toggle(false);
+    // Close the dropdown if:
+    // * An element in the shadow DOM gets focus (`relatedTarget` becomes `this` in this case)
+    // * An element outside this element gets focus
+    if (!this.contains(event.relatedTarget) || this === event.relatedTarget) {
+      this.open = false;
     }
   };
 
@@ -78,10 +81,9 @@ class BXDropdown extends HTMLElement {
     }
 
     if (event.key === 'Escape') {
-      // ensure the dropdown is closed
+      // ensure the dropdown is closed and focus the dropdown trigger
+      // we only want to force focus back on the trigger when the user presses `Escape`
       this.#toggle(false);
-      // focus the dropdown trigger - we only want to force focus back on the trigger when the user presses `Escape`
-      this.shadowRoot.querySelector(BXDropdown.triggerSelector).focus();
     }
   };
 
@@ -90,6 +92,8 @@ class BXDropdown extends HTMLElement {
     if (this.open) {
       const item = this.querySelector(BXDropdown.nonSelectedItemSelector);
       item.focus();
+    } else {
+      this.shadowRoot.querySelector(BXDropdown.triggerSelector).focus();
     }
   };
 
@@ -112,7 +116,7 @@ class BXDropdown extends HTMLElement {
       if (this.dispatchEvent(new CustomEvent(this.constructor.eventBeforeSelect, init))) {
         this.value = item.value;
         this.dispatchEvent(new CustomEvent(this.constructor.eventAfterSelect, init));
-        this.open = false;
+        this.#toggle(false);
       }
     }
```

asudoh:
@cal-smith Created [a merge commit ](https://github.com/carbon-design-system/carbon-custom-elements/pull/2/commits/37c119215a5146e6c6bca61e7c4a7683355fbe41) which I believe incorporates all your latest changes (plus a dedicated file for `FocusMixin`). Please feel free to back it out in case it diverges your intent. I haven't incorporated the changes I have suggested in this PR, but please let me know if you'd like. Thanks!

cal-smith:
Yeah that looks good ... thanks! I'll play around with the changes suggested ... originally I left `.open = newOpenState`, so I'll just swap it back to that :+1:

asudoh:
Looks that this is almost there @cal-smith! Do you want to migrate to `@HostListener` decorator? Guessing it may be a quick way to get around the TS error.

Another thing I'm wondering is if we can do `somethingSelector` vs. `selectorSomething` naming nit. The codebase takes the latter, but if you feel inclined to the former, I can go through the codebase to change them to the former. Please let me know - Thanks!

cal-smith:
> Looks that this is almost there @cal-smith! Do you want to migrate to `@HostListener` decorator? Guessing it may be a quick way to get around the TS error.

Yep! I've got some progress on that made in a local branch, I was just chasing down some interaction glitches

> Another thing I'm wondering is if we can do `somethingSelector` vs. `selectorSomething` naming nit. The codebase takes the latter, but if you feel inclined to the former, I can go through the codebase to change them to the former. Please let me know - Thanks!

Hmm, I've just been using `somethingSelector` instinctively ... I think it may flow a little better, but I'm not totally attached to it. I can flip everything to `selectorSomething` though

asudoh:
> Yep! I've got some progress on that made in a local branch, I was just chasing down some interaction glitches

Thanks @cal-smith! Looking forward to seeing the new version!

>> Another thing I'm wondering is if we can do `somethingSelector` vs. `selectorSomething` naming nit. The codebase takes the latter, but if you feel inclined to the former, I can go through the codebase to change them to the former. Please let me know - Thanks!

> Hmm, I've just been using `somethingSelector` instinctively ... I think it may flow a little better, but I'm not totally attached to it. I can flip everything to `selectorSomething` though

Thank you for sharing your thoughts! Let's go with `selectorSomething` for now, one reason is migration easiness from vanilla component options API.

cal-smith:
Okay cool! I'm happy with the interactions and patterns we're using here 👍 As far as I can tell we should be nicely compliant with all the usual dropdown/select a11y stuff.

src/components/dropdown/dropdown-item.ts
```diff
@@ -25,17 +51,45 @@ class BXDropdownItem extends LitElement {
     if (!this.hasAttribute('role')) {
       this.setAttribute('role', 'option');
     }
+    this._hKeydown = on(this, 'keydown', this._handleKeydown);
```
asudoh: Seems that `@HostListener()` can be used for here.

src/components/dropdown/dropdown-item.ts
```diff
@@ -8,7 +10,31 @@ const { prefix } = settings;
...
+  /**
+   * The handle for the `keydown` event on this element.
+   */
+  private _hKeydown: Handle | null = null;
+
+  /**
+   * Handles `keydown` event on this element.
+   */
+  private _handleKeydown = event => {
+    if (event.key === 'ArrowDown' || event.key === 'Down') {
```
asudoh: Utilization of a table used in structured list may make sense here: https://github.com/carbon-design-system/carbon-custom-elements/blob/c69001c/src/components/structured-list/structured-list-row.ts#L13-L21

src/components/dropdown/dropdown-item.ts
cal-smith: Thought I grabbed that one 😮 oops

src/components/dropdown/dropdown-item.ts
cal-smith: Ah yeah, something like that would work 🤔

PR #4, opened 2019-05-17
feat(components): move to lit-element

asudoh:
Going with `lit-element` and TypeScript. Some notes are:

* `src/typings` directory contains some type definitions for `carbon-components` utility code. Hoping that it gives better explanation with on those utility code with editors code assist feature.
* ~~The code imports styles as strings (as this repo has been doing since the beginning), but `lit-element` seems to expect [`CSSResult`](https://lit-element.polymer-project.org/api/classes/_lib_css_tag_.cssresult.html) as styles to set in Shadow DOM. The only way I have been able to find to convert between them is [`unsafeCSS()`](https://lit-element.polymer-project.org/api/modules/_lib_css_tag_.html#unsafecss) but hoping to find a better way in near future, probably coming up with a WebPack loader.~~ - Update: Created a [custom WebPack loader](https://github.com/carbon-design-system/carbon-custom-elements/blob/37b76683c9139ff4928edb851df414de4234d7a9/css-result-loader.js)
* I used to use `template()` as a method to return `lit-html` template, but in `lit-element` it's `render()`.
* Lots of other changes to align to `lit-element` API.
* Every properties/methods (except inherited one) are with TSDoc comments. Great if subsequent commits follows the same, so TypeScript's type annotation comes with good descriptions.
* Uses ESLint with `typescript-eslint` for linting. Still using Prettier for formatting.
* Hoping that the code has become a good TypeScript citizen, but @cal-smith please don't hesitate to suggest any better TypeScript usage!


src/index.ts
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable import/prefer-default-export */
+
+export { default as BXButton } from './components/dropdown/dropdown';
```
cal-smith: Did you mean `'./components/button/button'`?

asudoh: Oops good catch @cal-smith!

PR #8, opened 2019-05-24
chore(components): introduce @HostListener

asudoh:
`@HostListener(type, options?)` decorator attaches an event listener of the host element of a custom element.

The custom element class using `@HostListener()` has to extend `HostListenerMixin` that does the actual event listener work. Otherwise, `@HostListener()` throws an error.


cal-smith:
Sorry, not quite what I was meaning ... the old signature for the `HostListener` was amazing, I was thinking more for the class mixins - `HostListenerMixin` -> `@HostListenerMixin`

PR #15, opened 2019-06-07
chore(components): integrate icon helper

asudoh:
Also contains the following changes to `icon.ts`:

- Remove named export from `icon.ts` for the sake of the default export
- Prevent `<title>` content from being set as `<svg>`'s attribute
- Call `svg` tagged literal function directly so we don't have to call `unsafeHTML()`

@cal-smith Wondering if the changes to `icon.ts` makes sense to you... Please don't hesitate to speak up otherwise - Thanks!


src/components/icon/icon.ts
```diff
@@ -32,13 +31,12 @@ export const icon = (descriptor, attributes = {}) => {
...
-      content: [descriptor.attrs.title],
+      content: [title],
     });
     descriptor.attrs['aria-describedby'] = id;
   }
-  return svg`
-    ${unsafeHTML(toString(descriptor))}
-  `;
+  const strings = [toString(descriptor)];
+  return svg(Object.assign(strings, { raw: strings }) as TemplateStringsArray);
```
cal-smith: Ah  yeah, that makes a lot more sense :+1:

PR #18, opened 2019-06-08
chore(devenv): a11y change for Storybook preview

asudoh:
Does the following:

- Add `lang` attribute to `<html>`
- Let our a11y tool know that the focus-trap element is hidden


src/components/focus-trap/focus-trap.ts
```diff
@@ -0,0 +1,27 @@
+import settings from 'carbon-components/es/globals/js/settings';
+import { html, customElement, LitElement } from 'lit-element';
+import styles from './focus-trap.scss';
+
+const { prefix } = settings;
+
+@customElement(`${prefix}-focus-trap` as any)
+class BXFocusTrap extends LitElement {
```
cal-smith: Interesting ... I'm not sure I totally understand the function of this particular element ... is it primarily as a helper for storybook?

asudoh: Good question @cal-smith, it works the same manner as `<input class="bx--visually--hidden">` to ensure that `focus` event listener another component attached on document level fired when such component loses focus, but I see different ways to achieve the similar purpose (e.g. `@HostListener('blur')` we recently introduced). I removed `<bx-focus-trap>` given it makes things simpler.

asudoh: Oops I now recall another reason I had the focus-trap element; Lack of focus-trap element tends to cause viewport losing focus and at that point our code has no way to set the focus to the element we desire. (e.g. Focus-wrap for modal, though this library hasn't implemented it). Added it back.

cal-smith: Interesting ... So if I understand correctly, it's a utility component for the storybook environment ... would it make sense to keep it somewhere else? Maybe declare it in the `.storybook` folder? 

If it has utility in applications then for sure we'll want it here :+1:

asudoh: @cal-smith Makes sense (I was on the fence wrt publishing it vs. making it private) - Went ahead with the private route.

cal-smith: Cool, we can always re-evaluate if a need arises.

PR #22, opened 2019-06-17
feat(modal): introduce focus-wrap

asudoh:
This change does the following:

- Attempts to focus on the element in the following priority order, when modal gets open:
  - An element that application specifies via `data-modal-primary-focus` attribute
  - A primary button
  - The first focusable element in modal or modal itself
- Attempts to focus on the first focusable element in modal (or modal itself), when modal body loses focus, and the latest focused element follows the modal body in DOM order
- Attempts to focus on the last focusable element in modal (or modal itself), when modal body loses focus, and the latest focused element precedes the modal body in DOM order

This change also:

- Adds the selector for tabbable elements (borrowed from @cal-smith's #2), adding our custom elements that should be focusable
- Enhances the example focus-trap custom element to support skip-to-content feature

### DIFF
--- a/.storybook/components/focus-trap/focus-trap.scss
+++ b/.storybook/components/focus-trap/focus-trap.scss
@@ -1,5 +1,5 @@
 @import 'carbon-components/scss/globals/scss/css--helpers';
 
-:host(#{$prefix}-focus-trap) {
+:host(#{$prefix}-ce-focus-trap) {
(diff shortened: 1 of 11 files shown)


asudoh:
Thanks @cal-smith for reviewing. You are right, focus-wrap in modal is a big one in the keyboard nav topic.

PR #29, opened 2019-06-25
docs(project): i1 and i2 plans

asudoh:
Please don't hesitate to speak up if anybody have higher-priority items in your mind - Thanks!

### DIFF
--- a/plans/i1.md
+++ b/plans/i1.md
@@ -1,0 +1,8 @@
+# Iteration 1 (July 2019)
+
+- Continue understanding how lack of "type extension" (e.g. `<table is="bx-table">`) affects our library and how we can get around it
+- Tests to cover most components
+- Start developing some key components
+  - Form components (Checkbox/Number input/Text input/Text input)
+  - Combo box/Multi select
+  - Data table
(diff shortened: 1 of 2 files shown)


plans/i1.md
```diff
@@ -0,0 +1,8 @@
+# Iteration 1 (July 2019)
+
+- Continue understanding how lack of "type extension" (e.g. `<table is="bx-table">`) affects our library and how we can get around it
```
cal-smith: This is definitely a big one, especially WRT table ...

plans/i1.md
```diff
@@ -0,0 +1,8 @@
...
+- Tests to cover most components
+- Start developing some key components
+  - Form components (Checkbox/Number input/Text input/Text input)
+  - Combo box/Multi select
+  - Data table
```
cal-smith: Some of the problems with lack of extensions here can be sorted with some extra directives + a reasonable table model ... carbon angular for example, has a well defined model for table interactions (https://github.com/IBM/carbon-components-angular/blob/master/src/table/table-model.class.ts) ... something along those lines could help with our lack of native element extension

asudoh: Thank you for sharing the example! Your link is interesting for state management capability, given we need to provide at least a guidance there. Another thing we need to look into is if we can represent each table headers/rows/cells as components.

PR #37, opened 2019-07-08
feat(data-table): introduce data table

asudoh:
This change re-implements `carbon-components` style code temporarily for now, from the following reasons:

* Many of the style rules are not supported by Sass `@extend`
* There are some open questions in `carbon-components` style code, and fiddling with an alternate implementation may help solving those questions

Please see https://github.com/asudoh/carbon-custom-elements/blob/75cb84052984644ab4ad0bd0910982c5e6a8f686/src/components/data-table/README.md for the basic notion of how the data table in `carbon-custom-elements` is implemented. Please don't hesitate to bring up if there are any questions. Thanks!

### DIFF
--- a/src/components/data-table/README.md
+++ b/src/components/data-table/README.md
@@ -1,0 +1,240 @@
+# Data table
+
+Data table in `carbon-custom-elements` focuses on primitives that constructs table UI, consisting of the following:
+
+| Tag                      | Description     | HTML tag counterpart |
+| ------------------------ | --------------- | -------------------- |
+| `<bx-data-table>`        | The container   | N/A                  |
+| `<bx-table>`             | The table       | `<table>`            |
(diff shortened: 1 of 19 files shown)


asudoh:
Thank you @cal-smith for your thoughts! I focused on user interface primitices that achieved Carbon design of data table for now considering the complexity around the state management, especially with the combination of windowing/sorting and lazy-loaded data, but more thoughts on this topic would definitely be nice.

Wrt a11y, I looked for a way of keeping it accessible without using `<table>` given we cannot do `<table id="bx-table">`, and found https://www.w3.org/TR/wai-aria-practices/examples/table/table.html. Basically it's how the markup in this PR is structured.

cal-smith:
Ah interesting ... from what I can see there, as long as we apply the correct roles (we do) we can match the native `<table>` semantics ... awesome!

PR #39, opened 2019-07-10
chore(devenv): introdue theme switcher

asudoh:
This change introduces a theme switcher in our Storybook, consisting
three parts:

* The theme switcher UI, implemented as a Storybook add-on
* Sass code change to define CSS custom properties and to add a theme option to use CSS custom properties
* Build change to combine CSS code with CSS custom properties and one without it. The latter works as an escape hatch for IE

The theme switcher is roped off and requires an environment variable (`CARBON_CUSTOM_ELEMENTS_STORYBOOK_USE_CUSTOM_PROPERTIES`) to turn it on, given it incurs extra build time.


cal-smith:
This looks pretty sweet ✨ Just retrying the netlify build now

cal-smith:
So I got the preview to build, but I don't see the theme switcher? Is it not enabled yet?

asudoh:
My apologies @cal-smith that I forgot to mention that we need to set `true` to `CARBON_CUSTOM_ELEMENTS_STORYBOOK_USE_CUSTOM_PROPERTIES` environment variable to use the theme switcher, given theme switcher causes extra build time.

cal-smith:
Ah okay cool, I'll test it locally then later :+1:

PR #54, opened 2019-07-13
feat(multi-select): introduce multi select

### LINKED ISSUE #45: Component: Multi select
Probably start with some basic features.

asudoh:
Fixes #45.

Please refer to https://github.com/asudoh/carbon-custom-elements/blob/909e4d5b5ed6b757c4f47e5e26c04c0b437c2794/src/components/multi-select/README.md to see how it works.

### DIFF
--- a/src/components/multi-select/multi-select-item.ts
+++ b/src/components/multi-select/multi-select-item.ts
@@ -1,0 +1,80 @@
+import settings from 'carbon-components/es/globals/js/settings';
+import { ifDefined } from 'lit-html/directives/if-defined';
+import { html, property, customElement } from 'lit-element';
+import BXDropdownItem from '../dropdown/dropdown-item';
+import styles from './multi-select.scss';
+
+const { prefix } = settings;
+
+/**
+ * Multi select item.
+ */
+@customElement(`${prefix}-multi-select-item` as any)
+class BXMultiSelectItem extends BXDropdownItem {
+  /**
+   * Unique ID used for form elements.
+   */
+  protected _uniqueId = Math.random()
+    .toString(36)
+    .slice(2);
+
+  /**
+   * `true` if this multi select should be disabled. Corresponds to the attribute with the same name.
+   */
+  @property({ type: Boolean, reflect: true })
+  disabled = false;
+
+  /**
+   * The `name` attribute for the `<input>` for selection. Corresponds to `selection-name` attribute.
+   */
+  @property({ attribute: 'selection-name' })
+  selectionName = '';
+
+  connectedCallback() {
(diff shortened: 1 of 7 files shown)


src/components/multi-select/multi-select-item.ts
```diff
@@ -0,0 +1,98 @@
...
+import { ifDefined } from 'lit-html/directives/if-defined';
+import { html, property, customElement, LitElement } from 'lit-element';
+import styles from './multi-select.scss';
+
+const { prefix } = settings;
+
+/**
+ * Multi select item.
+ */
+@customElement(`${prefix}-multi-select-item` as any)
```
cal-smith: This name is kinda long (not that there's a reasonable alternative) ... but more specifically, I wonder if we could reuse `bx-dropdown-item` at all for multi-select ...

asudoh:
Great point @cal-smith, and you are right that drop down and multi select would share some code. To share more background - We have two dropdown codebase in our core library, one started with V6, another shares styling with multi select. The current `<bx-dropdown-item>` code uses the former, and because of that, we need to change `<bx-dropdown-item>` to move to the latter. Wrt whether we need specific `<bx-multi-select-item>` custom element, I'm on the fence, but here's the plan; We first move `<bx-dropdown-item>` to the markup that shares the styling with multi select, and then see how much common code we have with `<bx-multi-select-item>` in this PR. Wondering this strategy makes sense to you. Please don't hesitate to ping me if you have any questions and/or further concerns. Thanks!

P.S. Will look at the issues you found.

cal-smith:
Sure that strategy makes sense 👍 we may even find we can abstract most of the interactions away in some `bx-listbox` base components and keep the current interfaces on top.

Another question that came to mind - how does a screen reader handle reading out the visually active (but not focused) items? I haven't tested, but I'd be worried that it won't read out when you navigate to a new item.

asudoh:
> how does a screen reader handle reading out the visually active (but not focused) items?

Good question - Two approaches that came to my mind:

1. `aria-activedescendant`, seen at: https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-collapsible.html
2. Using an off-screen element that lets screen reader its contents as soon as the contents change (e.g. `<div class="bx--assistive-text" aria-live="assertive">`)

asudoh:
@cal-smith Updated this PR with a change to make `<bx-multi-select>` inherit `<bx-dropdown>` and `<bx-multi-select-item>` inherit `<bx-dropdown-item>`. As you have imagined, doing so reduced lots of duplicates.

Wrt the Firefox issue and the "clear all by keyboard" issue, I see both are gone with the new code. Please don't hesitate to bring up you find any issue or further questions/concerns in the code. Thanks!

PR #61, opened 2019-07-27
feat(modal): keyboard nav enhancements

asudoh:
This changes adds the following features to `<bx-modal>`:

* Introduces focus-pop behavior
* Traps focus by itself, for its focus-wrap feature

This change also adds a feature to `@HostListener()` decorator to hook event listeners on shadow root, which is useful if we want to get events that are not re-targeted.

### DIFF
--- a/src/components/modal/modal.ts
+++ b/src/components/modal/modal.ts
@@ -18,6 +18,11 @@ const FOLLOWING = Node.DOCUMENT_POSITION_FOLLOWING | Node.DOCUMENT_POSITION_CONT
 @customElement(`${prefix}-modal` as any)
 class BXModal extends HostListenerMixin(LitElement) {
   /**
+   * The element that had focus before this modal gets open.
+   */
+  private _launcher: Element | null = null;
+
+  /**
    * Handles `click` event on this element.
    * @param event The event.
    * @private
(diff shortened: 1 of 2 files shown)


src/components/modal/modal.ts
```diff
@@ -17,6 +17,11 @@ const FOLLOWING = Node.DOCUMENT_POSITION_FOLLOWING | Node.DOCUMENT_POSITION_CONT
 
 @customElement(`${prefix}-modal` as any)
 class BXModal extends HostListenerMixin(LitElement) {
+  /**
+   * The element that had focus before this modal gets open.
+   */
+  private _launcher: Element | null = null;
```
cal-smith: I _think_ you can just type this as `Element` and assign to null ...

src/globals/mixins/host-listener.ts
```diff
@@ -3,7 +3,7 @@ import on from 'carbon-components/es/globals/js/misc/on';
 /**
  * The format for the event name used by `@HostListener` decorator.
  */
-const EVENT_NAME_FORMAT = /^((document|window):)?(\w+)$/;
+const EVENT_NAME_FORMAT = /^((document|window|shadowRoot):)?(\w+)$/;
```
cal-smith: Neat! that's definitely a handy addition

src/components/modal/modal.ts
asudoh: Interesting point - Tried it, but got:

image

Looking for better way to write `null`-able...

cal-smith: Oh weird ... If we have `strictNullChecks` on that may be why ... `Element | null` would be the correct signature for null strictness 👍

asudoh: Uh I see - We have `strict` in `tsconfig.json` and that explains. Thank you for your thoughts here @cal-smith!

PR #67, opened 2019-07-31
feat(combo-box): introduce combo box

### LINKED ISSUE #44: Component: Combo box
Probably start with some basic features.

asudoh:
Fixes #44.


src/components/dropdown/dropdown.ts
```diff
@@ -506,15 +548,14 @@ class BXDropdown extends HostListenerMixin(FocusMixin(LitElement)) {
         <div
           role="button"
```
cal-smith: Just want to verify that `role="button"` is valid if the trigger isn't focussable ...

asudoh: Good point! Fixed.

asudoh:
Thank you for reviewing and finding the issues @cal-smith!

> - If I'm typing and press space it auto selects the first item

Couldn't reproduce it with a space key. Reproduced it with hitting backspace key to make the `<input>` empty and I guess this is what you meant, but please don't hesitate to correct me if I'm wrong here.

> - when I arrow key up and down the cursor moves to the start and end of the text, as well as moving the selection ... I'm not sure what's correct here, but that feels wrong

Makes sense, fixed.

> - when I clear the filter (with backspace) it doesn't always clear the selection right away (make a selection -> close the menu -> open the menu -> clear selection -> `x` still visible until tab out)

Great catch! Fixed.

cal-smith:
> Couldn't reproduce it with a space key. Reproduced it with hitting backspace key to make the <input> empty and I guess this is what you meant, but please don't hesitate to correct me if I'm wrong here.

I'm still seeing it ... the easiest way to replicate seems to be to start typing an option title, then press space, like so:

giphy

asudoh:
Got it - Thanks alot @cal-smith for clarifying! Fixed.

PR #70, opened 2019-08-02
feat(tile): introduce tile

asudoh:
This change also modified checkbox code to align with tile code.

### DIFF
--- a/src/components/checkbox/checkbox-story.ts
+++ b/src/components/checkbox/checkbox-story.ts
@@ -1,3 +1,4 @@
+import { ifDefined } from 'lit-html/directives/if-defined';
 import { html } from 'lit-element';
 import { storiesOf } from '@storybook/polymer';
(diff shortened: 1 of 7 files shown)


asudoh:
Good points, will add more variants as the next step. If you meant `<bx-tile>` working for all variants, I was on the fence on this, but took the current route given the possible complexity in underlying template. I can rename `<bx-tile>` to something like `<bx-simple-tile>` and make `<bx-tile>` a chooser for variant components if it's more desirable.

cal-smith:
More that the underlying template/carbon styles don't compose together super nicely (you can't really go tile -> selectable tile, they're effectively separate components) ... It's a good separation of concerns as is, so we shouldn't need to go with a chooser pattern.

asudoh:
Got it, thanks @cal-smith for clarifying!

PR #85, opened 2019-08-08
chore(devenv): add more React stories

asudoh:
This change adds stories of remaining components for React.

Refs #60.


src/components/overflow-menu/overflow-menu.ts
```diff
@@ -64,21 +64,22 @@ class BXOverflowMenu extends HostListenerMixin(LitElement) implements BXFloating
...
   }
 
-  attributeChangedCallback(name, old, current) {
-    if (old !== current) {
-      if (name === 'open' && !this._menuBody) {
+  createRenderRoot() {
+    return this.attachShadow({ mode: 'open', delegatesFocus: true });
+  }
+
+  updated(changedProperties) {
```
asudoh: This change was originally introduced for Angular which puts property binding the precedence, and thus tracking change in property rather than attribute.

asudoh:
@emyarod Rebased the code and eliminated redundant changes.

PR #87, opened 2019-08-12
feat(input): add input component 

### LINKED ISSUE #40: Component: Text input

cal-smith:
closes #40 

- adds `bx-input` component and related stories
- adds `bx-form-item` as the beginning of a set of form related components


cal-smith:
We'll probably want to hook it all up to the `ElementInternals` interface, but given that the label and element are just linked via `id` it should be as easy as generating and applying an `id` to the slotted input

cal-smith:
So `attachInternals` and friends aren't implemented in our polyfills yet (https://github.com/webcomponents/polyfills/issues/173), we'll have to wait on that ... from what I can see we can generally continue, and label->input linking should work as expected anyway ... we'll just be missing some more advanced form features for now (this is true for checkboxes and other wrapped native elements)

src/components/form/form-item.ts
```diff
@@ -0,0 +1,21 @@
...
+const { prefix } = settings;
+
+/**
+ * Presentational element for form items
+ */
+@customElement(`${prefix}-form-item`)
+export default class BXFormItem extends LitElement {
+  render() {
+    return html`
+      <div class="bx--form-item">
```
asudoh: To reduce the number of DOM nodes, would it make sense to make `:host(#{$prefix}-form-item)` inherit (`@extend`) `.#{$prefix}--form-item`?

src/components/input/input.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  /**
+   * Unique ID used for ID refs.
+   */
+  protected _id = Math.random()
+    .toString(36)
+    .slice(2);
+
+  render() {
+    const invalidIcon = WarningFilled16({ class: 'bx--text-input__invalid-icon' });
```
asudoh: Need to use `prefix`.

src/components/input/input.ts
```diff
@@ -0,0 +1,113 @@
...
+          name="${this.name}"
+          pattern="${this.pattern}"
+          placeholder="${this.placeholder}"
+          ?readonly="${this.readonly}"
+          ?required="${this.required}"
+          type="${this.type}"
+          value="${this.value}"
+        />
+      </div>
+      <div class="bx--form-requirement"><slot name="validation"></slot></div>
```
asudoh: Need to use `prefix`.

src/components/input/input.ts
```diff
@@ -0,0 +1,113 @@
...
+      <div class="${prefix}--text-input__field-wrapper" ?data-invalid="${this.invalid}">
+        ${this.invalid ? invalidIcon : null}
+        <input
+          ?autocomplete="${this.autocomplete}"
+          ?autofocus="${this.autofocus}"
+          class="${inputClasses}"
+          ?data-invalid="${this.invalid}"
+          ?disabled="${this.disabled}"
+          form="${this.form}"
+          id="${this._id}"
```
asudoh: Wondering if the ID there can be based on the host element ID, like https://github.com/carbon-design-system/carbon-custom-elements/blob/bc73ab5/src/components/dropdown/dropdown.ts#L88-L94. Alternatively, we can do something like just `input` given each shadow tree has unique ID namespace.

src/components/input/input.ts
```diff
@@ -0,0 +1,113 @@
...
+      <div class="${helperTextClasses}"><slot name="help-text"></slot></div>
+      <div class="${prefix}--text-input__field-wrapper" ?data-invalid="${this.invalid}">
+        ${this.invalid ? invalidIcon : null}
+        <input
+          ?autocomplete="${this.autocomplete}"
+          ?autofocus="${this.autofocus}"
+          class="${inputClasses}"
+          ?data-invalid="${this.invalid}"
+          ?disabled="${this.disabled}"
+          form="${this.form}"
```
asudoh: Wondering if this attribute actually works, given the `<input>` here is in shadow DOM and the `<form>` would be in light DOM or another shadow DOM - Where the namespace of ID is different. Good attempt to cover possible attribute BTW!

src/components/input/input.scss
```diff
@@ -0,0 +1,7 @@
+$css--plex: true !default;
+
+@import 'carbon-components/scss/components/text-input/text-input';
+
+:host(#{$prefix}-input) {
+  width: 100%;
```
asudoh: Just wondering; Where does the `100%` here came from?

src/index.ts
```diff
@@ -1,13 +1,17 @@
 export { default as BXButton } from './components/button/button';
-export { default as BXTable } from './components/data-table/table';
```
asudoh: Was the removals in this file intentional, or did something happen with rebasing...?

src/components/input/input.ts
asudoh: Yeah dropping `form` attribute for now and TBD-planning for future makes sense for now 👍

src/components/input/input.scss
asudoh: Got it 💯 - Thank you @cal-smith for explaining this!

src/index.ts
asudoh: OK do you want to fix this? I can do manual merge here if you'd like - Let me know. Thanks!

cal-smith: I don't _think_ there's anything to fix? Let me know if there's an export that's missing though!

cal-smith:
Ready for another review @asudoh!

src/index.ts
asudoh: Uh - I now see what you meant. My apologies @cal-smith! In case if you can do a favor - Can the sorting be based on directory name then file name, instead of the import name? In this way we can avoid a condition e.g. UI shell imports being split to two places.

src/components/input/input.ts
```diff
@@ -0,0 +1,168 @@
...
+    const labelClasses = classnames(`${prefix}--label`, {
+      [`${prefix}--label--disabled`]: this.disabled,
+    });
+
+    const helperTextClasses = classnames(`${prefix}--form__helper-text`, {
+      [`${prefix}--form__helper-text--disabled`]: this.disabled,
+    });
+
+    return html`
+      <label class="${labelClasses}" for="${this._uniqueId}">
```
asudoh:
```suggestion
      <label class="${labelClasses}" for="${this._inputId}">
```

src/components/input/input.ts
```diff
@@ -0,0 +1,168 @@
...
+   * The helper text. Corresponds to `helper-text` attribute.
+   */
+  @property({ attribute: 'helper-text' })
+  helperText = '';
+
+  @property()
+  id = '';
+
+  @property({ type: Boolean, reflect: true })
+  invalid = false;
```
asudoh: Discussion; Reflecting the discussion we had in React codebase, this codebase stopped supporting "empty validity message" scenario and thus `--invalid` class is added based on `validityMessage` being non-empty (e.g. https://github.com/carbon-design-system/carbon-custom-elements/blob/29ad0c9/src/components/dropdown/dropdown.ts#L481), instead of using another property. Wondering if this make sense to you or you think separate `invalid` is more desirable despite the discussion of "empty validity message". What do you think?

src/components/input/input.ts
```diff
@@ -0,0 +1,168 @@
...
+      <div class="${prefix}--text-input__field-wrapper" ?data-invalid="${this.invalid}">
+        ${this.invalid ? invalidIcon : null}
+        <input
+          ?autocomplete="${this.autocomplete}"
+          ?autofocus="${this.autofocus}"
+          class="${inputClasses}"
+          ?data-invalid="${this.invalid}"
+          ?disabled="${this.disabled}"
+          id="${this._inputId}"
+          invalid="${this.invalid}"
```
asudoh: Couldn't find `invalid` as a native property of `<input>`, so wondering if you have any intent behind `invalid` attribute here...?

src/components/input/input.ts
```diff
@@ -0,0 +1,168 @@
...
+ */
+@customElement(`${prefix}-input`)
+export default class BXInput extends LitElement {
+  /**
+   * May be any of the standard HTML autocomplete options
+   */
+  @property()
+  autocomplete = '';
+
+  @property({ type: Boolean })
```
asudoh: It'd be nice if we can add comments to properties, even if it came from native `<input>` - Given `BXInput` is not technically inheriting `<input>` (due to lack of type extension), rather it's a wrapper.

src/components/input/input.ts
cal-smith: Huh, I could swear that was an attribute ... My bad!

src/components/input/input.ts
cal-smith: Since the invalid message can be an element `slot`ted in to the component, it seems to make more sense to keep the simpler `invalid` toggle instead of requiring a full DOM insertion.

We should be able to check for the existence of a `slot`ted element, or a valid `validityMessage` and only then apply the invalid state - that way we can avoid the whole "empty validity message" issue (if that's the design we're set on)

src/index.ts
cal-smith: For sure, my bad if I messed it all up 😅 I'll do a more nuanced sort

src/components/input/input.ts
asudoh: Good point on `@slotchange`! Though I couldn't rattle off a record that led to in favor of reduced properties (attributes), this is what I recall (and thus this codebase took such approach for now). Another bonus here is that introducing `invalid` attribute/prop (back) won't be a breaking change (though the opposite is a breaking change and thus React codebase still has `invalid` prop).

src/components/input/input.ts
asudoh: 👍 No problem at all!

src/index.ts
asudoh: Not a problem @cal-smith! It's all my bad missing where the deleted portions went 🙇

cal-smith: Should be sorted a lot better now :+1:

cal-smith:
> Is it possible for you to support automatic invalid state detection by existence of validity message later in a separate PR?

I _can_ I'm just not sure that API makes the most sense from a webcomponents perspective ... If I'm using the component in a non-lit-element context, it's a lot more work to add slotted content vs. simply toggling an attribute. It also adds another location for content to be generated and maintained, vs. having it all live in the DOM

asudoh:
Sounds fair, DM'ing @cal-smith for a bit of clarification, but despite the topic the change looks good! Thanks again @cal-smith for finishing this up!

src/components/input/input.ts
asudoh: Update: We are keeping both `invalid` and `validity-message`, at least for now, given there is not much connection in the meaning of those attributes.

Issue #46, opened 2019-07-10
Component: Data table


asudoh:
Basic feature have been implemented in #37. The remainders may go to one of following milestones.

PR #97, opened 2019-08-19
feat(search): introduce search box

asudoh:
Refs #46.


asudoh:
Thanks @cal-smith for reviewing! https://deploy-preview-97--distracted-euler-061082.netlify.com/?path=/story/search--default seems to be working now, but please let me know if you are still seeing otherwise!

PR #98, opened 2019-08-19
chore(build): introduce d.ts files

### LINKED ISSUE #90: Build: Consider shipping .d.ts files
For TypeScript users. Given our source code is written in TypeScript, likely that we can just use the appropriate toolchain for that purpose.

asudoh:
Fixes #90.


asudoh:
Thank you for reviewing @cal-smith!

PR #99, opened 2019-08-20
feat(inline-loading): introduce inline loading

asudoh:


asudoh:
@cal-smith Sorry for the breakage. It was broken by Babel preset for Vue getting in place just by installing `@storybook/vue`. Put [a patch](https://github.com/carbon-design-system/carbon-custom-elements/commit/214121c2189cc37ad344902f5b3addf1b2ebef96) to `master` and merged it to here, and the problem seems to have been solved.

Issue #91, opened 2019-08-14
Build: Consider auto-generating React wrapper


asudoh:
Currently (most of) the React stories (introduced in #60) have wrapper React components to correctly map React props to attributes/events. This issue is for tracking an effort to potentially auto-generate the wrapper using the content of `@property()` decorator as well as one of `eventSomething` static properties.

PR #100, opened 2019-08-21
chore(devenv): use auto-generated React wrapper

asudoh:
This change converts all React stories to the auto-generated React wrapper introduced in #95.

This change also adds support for components extending another component to the Babel plugin that generates the React wrappers.

Refs #91.


asudoh:
@emyarod Fixed the Netlify timeout issue. As always, please don't hesitate to ping me if you have any questions. Thanks!

PR #132, opened 2019-09-27
feat(date-picker): introduce date picker

asudoh:


asudoh:
Good catch @cal-smith! Fixed by adding the knob for `placeholder`.

PR #133, opened 2019-09-29
feat(slider): introduce bx-slider

asudoh:


asudoh:
Good point @cal-smith! Here's the code does:

1. It does _not_ set the value to `0` as soon as user enters a negative value
2. But we ensure the thumb position goes to the point of `0`
3. When user moves the thumb it keeps in `0` .. `100` range

Wondering if it makes sense...? Given `-194` there is what user explicitly entered, I'm guessing that we'd show that as invalid.

cal-smith:
Sorry, I wasn't clear, the `-194` was from me dragging the mouse around 😮 I was surprised because I did see some bounds checking in the code.

We definitely shouldn't do any bounds checking on the manual input - that's really up to the app to decide (they may show invalid, or maybe they'll just correct the user)

asudoh:
My apologies for misinterpreting your earlier comment @cal-smith! Great catch and fixed.

asudoh:
🙇 thanks alot @cal-smith for reviewing!

PR #138, opened 2019-10-03
chore(components): remove observer of ID attribute

asudoh:
This change is a continuation of #137, removing redundant code observing `id` attribute for unique-ID generation purpose.

As stated in #137, auto-generated unique ID is no longer needed at most places, given shadow DOM creates its own ID space.


cal-smith:
I suppose this means we can drop the `id` property everywhere else too 🤔

asudoh:
@cal-smith Good point, I think yes, public `id` property is no longer needed, at least for most of the components.

PR #150, opened 2019-10-15
feat(tooltip): introduce bx-tooltip-definition

asudoh:
Also `bx-tooltip-icon`.


asudoh:
Thank you for catching that @emyarod! Fixed.

PR #156, opened 2019-10-23
feat(angular): introduce form directives

asudoh:
This change implements Angular directives that implement `ControlValueAccessor` interface for components that should be able to interact with `<form>` (e.g. `<bx-input>`), so `[(ngModel)]` works with those.

This change relies on automated test cases for testing and manual test case will be in a separate PR. You can take a glance at manual test case [here](https://github.com/asudoh/carbon-custom-elements/tree/form-examples/examples/codesandbox/form/angular), which can be run by the following:

```sh
> git fetch git@github.com:asudoh/carbon-custom-elements.git form-data
> git checkout FETCH_HEAD
> gulp clean
> gulp build
> git fetch git@github.com:asudoh/carbon-custom-elements.git form-examples
> git checkout FETCH_HEAD
> cd examples/codesandbox/form/angular
> yarn install
> cd node_modules/carbon-custom-elements
> rm -Rf es
> cp -r ../../../../../../es .
> cd ../..
> yarn start
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+dist
 docs/js
 es
 node_modules
--- a/gulp-tasks/build.js
+++ b/gulp-tasks/build.js
@@ -107,17 +107,6 @@ module.exports = {
       );
     },
 
-    angular() {
-      const tsProject = typescript.createProject(path.resolve(__dirname, '../tsconfig.json'));
-      const { js } = gulp
-        // Ensures type references are in the TS project by adding `.d.ts` here
-        .src([`${config.srcDir}/directives-angular/**/*.ts`, `${config.srcDir}/**/*.d.ts`], { base: config.srcDir })
-        .pipe(plumber())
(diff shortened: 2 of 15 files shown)


asudoh:
Thank you for your feedback @cal-smith! After some experiments with `ng-packagr`, I'm inclined to withdrawing this PR, given `ng-packagr` seems to generate a package with several Angular-specific contents and I'm not sure if it's a good idea to ship such generated package or merge our package contents in there.

An immediate approach we can take is providing a documentation with the directives required to use our form components with Angular (that is one I took before I came up with an idea of this PR). In future, we can consider isolating our Angular directive code (in a separate repo or in a monorepo structure) so we can provide more OOTB experience. What do you think?

asudoh:
As we discussed further offline, closing this PR for now. @cal-smith told me there is advanced API in `ng-packagr` that we may investigate.

PR #160, opened 2019-11-02
fix(textarea): reflect value attribute on init

asudoh:
This change ensures that `value` attribute of `<bx-textarea>` is reflected to the `<textarea>` in shadow DOM at initialization, where `lit-element` attempts to reflect `value` attribute to `value` property before it renders `<textarea>`.


asudoh:
Thank you for reviewing @cal-smith! The problem happens when `lit-element` attempts to sync `value` attribute to `value` prop at initialization, when the `<tetarea>` hasn't been rendered yet (and thus `this._textarea` is `null`).

PR #161, opened 2019-11-06
test(react-wrapper): add tests

asudoh:


tests/spec/createReactCustomElementType_spec.ts
```diff
@@ -0,0 +1,148 @@
...
+  let container;
+  const events = new EventManager();
+
+  beforeEach(function() {
+    container = document.createElement('div');
+    document.body.appendChild(container);
+  });
+
+  describe('Attribute mapping', function() {
+    it('Should use the property name the attribute name by default', function() {
```
emyarod: can this be reworded? it's a bit confusing

asudoh: Thanks @emyarod for your thoughts! How about adding "as"...? (...use the property name as the attribute name...)

emyarod: yes that would be better

asudoh: Thanks @emyarod! Updated.

PR #165, opened 2019-11-08
feat(angular): introduce form directives

asudoh:
(Continuation of #156. Difference from the latest code of #156 is that this PR does APF support by using `ngc`, a primitive behind `ng-packagr`)

This change implements Angular directives that implement `ControlValueAccessor` interface for components that should be able to interact with `<form>` (e.g. `<bx-input>`), so `[(ngModel)]` works with those.

This change relies on automated test cases for testing and manual test case will be in a separate PR. You can take a glance at manual test case [here](https://github.com/asudoh/carbon-custom-elements/tree/form-examples/examples/codesandbox/form/angular), which can be run by the following:

```sh
> git fetch git@github.com:asudoh/carbon-custom-elements.git form-data
> git checkout FETCH_HEAD
> gulp clean
> gulp build
> git fetch git@github.com:asudoh/carbon-custom-elements.git form-examples
> git checkout FETCH_HEAD
> cd examples/codesandbox/form/angular
> yarn install
> cd node_modules/carbon-custom-elements
> rm -Rf es
> cp -r ../../../../../../es .
> cd ../..
> yarn start
```

### DIFF
--- a/src/directives-angular/slider.ts
+++ b/src/directives-angular/slider.ts
@@ -1,0 +1,31 @@
+/**
+ * @license
+ *
+ * Copyright IBM Corp. 2019
+ *
+ * This source code is licensed under the Apache-2.0 license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import { Directive, forwardRef } from '@angular/core';
+import { NG_VALUE_ACCESSOR, NumberValueAccessor } from '@angular/forms';
+import settings from 'carbon-components/es/globals/js/settings';
+import BXSlider from '../components/slider/slider';
+
+const { prefix } = settings;
+
+@Directive({
+  selector: `${prefix}-slider[formControlName],${prefix}-slider[formControl],${prefix}-slider[ngModel]`,
+  host: {
+    [`(${BXSlider.eventAfterChange})`]: 'onChange($event.detail.value)',
+    '(blur)': 'onTouched()',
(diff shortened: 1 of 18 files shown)


src/directives-angular/slider.ts
```diff
@@ -0,0 +1,35 @@
...
+  '(blur)': 'onTouched()',
+};
+
+// NOTE: Referring `BXSlider.eventAfterChange` seems to cause ng-packagr to package up `src/components/slider.ts` code,
+// Which is not desirable
+host[`(${prefix}-slider-changed)`] = 'onChange($event.detail.value)';
+
+@Directive({
+  selector: `${prefix}-slider[formControlName],${prefix}-slider[formControl],${prefix}-slider[ngModel]`,
+  host,
```
cal-smith: If we're going to use the `host` property on the `Directive` decorator, we should probably make sure we're consistent with our usage 👍

asudoh: Talked offline with @cal-smith explaining that this inconsistency came from a workaround for a (presumable) limitation in `ngc` which doesn't seem to understand computed property names in object literal, like:

```javascript
host: {
  ...
  [`(${prefix}-slider-changed)`]: 'onChange($event.detail.value)'
  ...
}
```

Thanks @cal-smith for sharing your thoughts!

PR #166, opened 2019-11-08
feat(code-snippet): introduce code snippet

asudoh:


src/components/code-snippet/code-snippet.ts
```diff
@@ -0,0 +1,270 @@
...
+      _showCopyButtonFeedback: showCopyButtonFeedback,
+      _showExpando: showExpando,
+      _handleClickCopyButton: handleClickCopyButton,
+      _handleClickExpando: handleClickExpando,
+      _handleSlotChange: handleSlotChange,
+    } = this;
+    // Ensures no extra whitespace text node
+    // prettier-ignore
+    return html`
+      ${type === CODE_SNIPPET_TYPE.INLINE
```
cal-smith: Would it maybe make sense to instead use multiple `return`s instead of bundling everything into one `html` template?

cal-smith:
Looks good overall 👍

src/components/code-snippet/code-snippet.ts
asudoh: Makes sense, thank you for your feedback! - Updated.

PR #168, opened 2019-11-12
chore(components): use TypeScript 3.7 features

asudoh:
This change upgrades TypeScript to 3.7, which allows us to use nullish coalescing (`??`) and optional chaining (`?.`) stage-3 features.

This change also includes a change to `src/coding-conventions.md` that guides developers determine when to use those operators.


asudoh:
Oops `ngc` doesn't seem to have supported TypeScript 3.7 yet... Closing for now, will reopen once it happens.

asudoh:
Temporary disabling TS version check with `ngc` - Reopening this PR. @cal-smith Could you please re-review? Thanks!

src/coding-conventions.md
```diff
@@ -186,8 +186,11 @@ To avoid problems with collation, the primary means for user to determine order
 
 If you get TypeScript "may be null" errors, think twice to see if there is such edge case:
 
-- If so, do such check to throw more reasonable exception or to make it no-op if the condition is not met.
-- If not, you can now add non-null assertion operator (`!`) - But again, don't do that blindly.
+- If some other portion of your code ensures the `null` condition won't happen and somebody else won't likely break that, use non-null assertion operator (`!`) - But don't do that blindly.
```
emyarod:
```suggestion
- If some other portion of your code ensures the `null` condition won't happen and nothing else is likely to break it, use the non-null assertion operator (`!`) - But don't blindly do so.
```

src/coding-conventions.md
```diff
@@ -186,8 +186,11 @@ To avoid problems with collation, the primary means for user to determine order
...
+- Otherwise, add code to do `null` check to do one of the following:
```
emyarod:
```suggestion
- Otherwise, add code to perform a `null` check by doing one of the following:
```

src/coding-conventions.md
```diff
@@ -186,8 +186,11 @@ To avoid problems with collation, the primary means for user to determine order
...
+  - Throw an exception that well explains why the `null` value won't be acceptable and (if applicable) what mistake may cause that wrong condition
```
emyarod:
```suggestion
  - Throw an exception that explains why the `null` value won't be acceptable and (if applicable) what mistake may cause that wrong condition
```

asudoh:
Merging, @cal-smith please don't hesitate to bring up if you find anything that should be rectified. Thanks!

tsconfig-angular-esm5.json
```diff
@@ -7,6 +7,7 @@
   },
   "angularCompilerOptions": {
     "annotateForClosureCompiler": true,
+    "disableTypeScriptVersionCheck": true,
```
cal-smith: Interesting! I didn't know about this option

PR #172, opened 2019-11-13
chore(dependencies): upgrade node-sass

asudoh:
This change upgrades `node-sass` to `4.12.x` or higher, that supports Node 12.


asudoh:
Yeah we should consider switch - Thanks @cal-smith for reviewing!

PR #187, opened 2019-11-27
docs(examples): update TypeScript 3.7 for Angular

asudoh:
This change updates Angular examples to make it work with `carbon-custom-elements` version compiled with TypeScript 3.7.

Refs #168.


asudoh:
After explaining where the error came from (`.d.ts` of our component code) and the roadmap behind Angular and this library (e.g. when Angular10 is likely to be released), we end up with making Angular9 the minimum supported version. Thank you @cal-smith for all your feedback!

PR #191, opened 2019-11-28
refactor(tag): remove unused tag selectors

emyarod:
This PR targets the host elements for styles and removes unused selectors from the stylesheet


src/components/tag/filter-tag.ts
```diff
@@ -23,7 +24,22 @@ const { prefix } = settings;
...
+  protected _buttonNode!: HTMLButtonElement;
+
+  /**
+   * Handles `click` event on this element.
+   * @param event The event.
+   */
+  @HostListener('shadowRoot:click')
+  // @ts-ignore: The decorator refers to this method but TS thinks this method is not referred to
+  private _handleClick = (event: MouseEvent) => {
+    if (event.composedPath().indexOf(this._buttonNode!) >= 0) {
```
asudoh: Sorry probably it was from my example code... Did you mean you didn't want the event to be fired only when this custom element is disabled, is it correct?

Side note (not related to this PR): Wondering if a UX has been defined for "closed" (hidden) state.

asudoh:
```suggestion
    if (this.disabled && event.composedPath().indexOf(this._buttonNode!) >= 0) {
```

Just in case this makes my above comment clearer. Thanks again for your fix @emyarod!

PR #196, opened 2019-12-01
fix(pagination): fix for plural

asudoh:
This change makes `formatStatusWithDeterminateTotal` property more compatible with `i18next`, where `count` is the only property that `i18next` uses for switching between singular/plural in the string interpolation function.


src/components/pagination/pageination-story.ts
```diff
@@ -30,7 +31,11 @@ export const defaultStory = ({ parameters }) => {
         <option value="20">20</option>
         <option value="30">30</option>
       </bx-page-sizes-select>
-      <bx-pages-select></bx-pages-select>
+      ${total == null
```
emyarod: would this work instead of ternary?

```js
(total === null) && html`<bx-pages-select></bx-pages-select>`
```

asudoh: I think `lit-html` renders `false` if I understand it correctly. Good thought to make it more idiomatic, though.

PR #200, opened 2019-12-04
feat(data-table): introduce table toolbar

asudoh:
This change introduces table toolbar, with search (filtering) UI and its demonstration.


asudoh:
Thank you for reviewing @cal-smith! And great point about the menu position upon window resizing. Created: #213

PR #201, opened 2019-12-05
chore(build): support RTL

asudoh:
This change adds the following:

* A build generate `*.rtl.css.js`, which is a RTL version of `CSSResult` module
* An environment variable for our dev env (Storybook), `STORYBOOK_CARBON_CUSTOM_ELEMENTS_USE_RTL`, to demonstrate/test RTL there

### DIFF
--- a/.storybook/angular/config.ts
+++ b/.storybook/angular/config.ts
@@ -19,6 +19,10 @@ import theme from './theme';
 import DocsPage from '../DocsPage';
 import containerStyles from '../_container.scss'; // eslint-disable-line import/first
 
+if (process.env.STORYBOOK_CARBON_CUSTOM_ELEMENTS_USE_RTL === 'true') {
+  document.documentElement.setAttribute('dir', 'rtl');
+}
+
 addParameters({
   docs: {
     container: DocsContainer,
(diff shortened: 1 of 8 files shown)


asudoh:
@cal-smith thank you for your thoughts here - Here's something I shared offline:

Single CSS build supporting LTR/RTL is something I considered, but didn’t do it (for now) for a couple of reasons:

* Performance: Having CSS supporting both doubles the CSS content for any rules with left/right (or their equivalents)
* [`:dir` pseudo class](https://developer.mozilla.org/en-US/docs/Web/CSS/:dir): Currently Firefox is the only browser that allows to auto-detect in CSS if an element is RTL and apply RTL version of styles for them. A workaround is using `dir` attribute, but it doesn’t work with style inheritances of nested shadow DOM.

I think it’s a natural thought that a single CSS supporting LTR/RTL is nice, though. That’s one of the reasons why I thought [shipping the Sass source](https://github.com/carbon-design-system/carbon-custom-elements/pull/164) is needed. Once the standards evolve that makes single build more sensible, application can create such “hybrid” build from the Sass source.

BTW I should have brought up that the change uses https://rtlcss.com behind the scene, and thus subsequent changes have directives for that tool. And the all subsequent changes (including one for slider) depends on #201 to see how it works in action.

PR #208, opened 2019-12-08
chore(dependencies): update peerDependencies

asudoh:


asudoh:
Closing, `lodash.debounce` was not needed as a peer-dependency, at least for now.

PR #214, opened 2019-12-11
fix(date-picker): support changing date format

asudoh:
This change allows change in `date-format` attribute after initialization reflected correctly to the underlying Flatpickr.


src/components/date-picker/date-picker.ts
```diff
@@ -259,7 +259,7 @@ class BXDatePicker extends LitElement {
   calendar: FlatpickrInstance | null = null;
 
   /**
-   * The date format to use. Corresponds to `date-format` attribute.
+   * The frdate format to let Flatpickr use. Corresponds to `date-format` attribute.
```
cal-smith:
```suggestion
   * The date format to let Flatpickr use. Corresponds to `date-format` attribute.
```

Unless `frdate` is a specific kind of format?

asudoh: My apologies 🙇 I think it simply was a typo.

PR #215, opened 2019-12-11
fix(floating-menu): catch up with resizing

### LINKED ISSUE #213: Floating menu: Cope with window resize
Current floating menu code does not have a mechanism to respond to window resize, which needs to be addressed.

asudoh:
This change ensures that floating menu is kept at the right position upon element resizing, by having floating menu observe resize of the floating menu container as well as the parent node of the trigger button.

Fixes #213.


asudoh:
Thank you for reviewing @cal-smith! - I think the closing thing is a separate issue, let's track it at: #228

cal-smith:
Sounds good 👍

PR #219, opened 2019-12-11
chore(component): change classnames by classMap

### LINKED ISSUE #218: Replace classnames by lit-html/classMap
We tested that the last PR that @asudoh did about the babel dependencies and ESM modules it worked (really thanks!) but we found that we continue having problems with the `classnames` library now. This library use CJS modules so our proposal is to change this library by the `classMap` imported by the library `lit-html` that you have now as dependency in the project. With this change and taking the `button` component as an example the next code:
```javascript
const classes = classnames(`${prefix}--btn`, {
      [`${prefix}--btn--${kind}`]: kind,
      [`${prefix}--btn--disabled`]: disabled,
      [`${prefix}--btn--icon-only`]: hasIcon && !hasMainContent,
      [`${prefix}--btn--sm`]: small,
      [`${prefix}-ce--btn--has-icon`]: hasIcon,
    });
```
would change to this:
```javascript
const classes = classMap({
          [`${prefix}--btn`]: true,
          [`${prefix}--btn--${kind}`]: kind,
          [`${prefix}--btn--disabled`]: disabled,
          [`${prefix}--btn--icon-only`]: hasIcon && !hasMainContent,
          [`${prefix}--btn--sm`]: small,
          [`${prefix}-ce--btn--has-icon`]: hasIcon
        });
```
For the `button` component at least this worked locally in the fork that I did so if you are agree I can try to make a PR with this change in the different components.

Tansito:
Remove classnames dependency to avoid CJS and changed the functionality by classMap.

Fixes #218


Tansito:
@asudoh I am going to ask you for help in one of the tests because I mostly could fix everything but the next test:

Screen Shot 2019-12-11 at 6 37 35 PM

Is giving me problems. If the `classMap` is empty shows an empty property `class` so that's correct and I updated the snapshot related with the `radio-button` but It continues failing in that test.

I hope this PR could be useful for you too.

abdonrd:
I want to help here with https://github.com/carbon-design-system/carbon-custom-elements/pull/220 😄

src/components/modal/modal.ts
```diff
@@ -150,8 +150,15 @@ class BXModal extends HostListenerMixin(LitElement) {
   open = false;
 
   render() {
-    const containerClasses = classnames(`${prefix}--modal-container`, {
-      [this.containerClass]: this.containerClass,
+    const _containerClass = {};
+    if (this.containerClass) {
+      this.containerClass.split(' ').forEach(externalClass => {
+        _containerClass[externalClass] = true;
```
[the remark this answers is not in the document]
Tansito: Sure!

src/components/modal/modal.ts
```diff
@@ -150,8 +150,10 @@ class BXModal extends HostListenerMixin(LitElement) {
   open = false;
 
   render() {
-    const containerClasses = classnames(`${prefix}--modal-container`, {
-      [this.containerClass]: this.containerClass,
+    const containerClass = this.containerClass.split(' ').reduce((acc, item) => ({ ...acc, [item]: true }), {});
```
asudoh:
```suggestion
    const containerClass = this.containerClass.split(' ').filter(Boolean).reduce((acc, item) => ({ ...acc, [item]: true }), {});
```

My apologies for not suggesting this earlier 🙇

Tansito: Oh no please, thank you your solution is by far better than what I was doing hehe

Tansito:
Sorry @asudoh I will need your help probably with the last issue in the tests. I was trying to reproduce it locally but in my branch is working correctly and pass all the tests so I couldn't figure out exactly what is happening.

PR #223, opened 2019-12-12
chore: replace lodash with lodash-es

### LINKED ISSUE #221: Dependencies: Consider switching to lodash-es
`<bx-modal>` depends on `lodash.findlast` and `<bx-slider>` depends on `lodash.throttle`. Neither of them has ESM support. Switching to `lodash-es` would pave the cowpath to native ESM support as standards evolve, e.g. [Import Maps](https://wicg.github.io/import-maps/).

abdonrd:
Fixes #221.


asudoh:
Thanks @abdonrd for jumping into this quickly! Not sure if there is, but if there is TS definition for `lodash-es` I think it'll fix the CI in part of the whole.

abdonrd:
@asudoh I add the types to all the `package.json`s, but it seems that it is not enough.

asudoh:
@abdonrd I think there are several type mismatches, probably one in the code vs. one in `@types/lodash-es`. Will take a look tomorrow.

asudoh:
* `data-table-story*.ts`: Seems that `@types/lodash-es` assumes that the function to be debounced takes no argument. `debounce(funcToBeDebounced as () => void), duration)` seems to help
* `slider.ts`: Changing [type definition of `_throttledHandleMousemoveImpl`](https://github.com/carbon-design-system/carbon-custom-elements/blob/v0.2.0/src/components/slider/slider.ts#L57) to use `MouseEvent` instead of `Event` seems to fix the problem
* `date-picker_spec.ts`: [Type definition of `appendTo` Flatpickr configuration](https://github.com/flatpickr/flatpickr/blob/v4.6.3/src/types/options.ts#L264) does not seem to take `null`. Adding `as HTMLElement` or `?? undefined` to `datePicker!.shadowRoot!.getElementById('floating-menu-container')` seems to help

.storybook/vue/create-vue-bindings-from-props.ts
```diff
@@ -7,7 +7,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import pickBy from 'lodash.pickby';
+import pickBy from 'lodash-es/pickby';
```
asudoh: Forgot to mention this thing, my apologies 🙇 

```suggestion
import pickBy from 'lodash-es/pickBy';
```

abdonrd: My fault! haha Updated!

PR #230, opened 2019-12-16
fix(number-input): fix for React

asudoh:
This change makes `<bx-number-input>` code support the Babel plugin that generates its React wrapper, by putting the getters of declared properties before their setters.

This change also ensures that failures in such Babel plugin causes CI error.

My apologies for not catching this earlier @cal-smith 🙇


asudoh:
My apologies again @cal-smith - It isn't documented and the Babel plugin may better handle setter-only declared properties. Just noticed the build error (the Babel plugin threw) when I was doing integration tests of CodeSandbox examples with the latest code.

PR #245, opened 2019-12-19
fix(floating-menu): add focus management

### LINKED ISSUE #228: Overflow menu and interactive tooltip: Problems with closing
From @cal-smith at #215:

> I will note that the overflow menu and tooltip don't seem to close properly - the tooltip sometimes closes when I click on it, but doesn't seem to close when I click outside it, while the overflow menu never seems to close (clicking the trigger, or outside don't have any affect) ... I don't see any errors in the console unfortunately, so I'm not sure where the cause is

asudoh:
This change adds close-on-click-outside feature to floating menus, by adding close-on-blur feature. This change also moves focus to floating menu body when floating menu gets open by user gesture.

Fixes #228.


asudoh:
Great catch @cal-smith! Fixed.

asudoh:
It seem to have been an old bug, but good catch @cal-smith! Fixed.

PR #288, opened 2020-01-26
chore(devenv): update license checker

asudoh:


asudoh:
Good idea @cal-smith - Yes something we should consider.

PR #301, opened 2020-01-31
chore(build): settings for CF deploy

asudoh:


manifest.yml
```diff
@@ -0,0 +1,7 @@
+---
+applications:
+- name: carbon-custom-elements
+  memory: 64M
+  buildpack: https://github.com/cloudfoundry/staticfile-buildpack.git
+routes:
+  - route: custom-elements.carbondesignsystem.com
```
cal-smith: Our own subdomain!

We can also use `github-pages` for this (`carbon-components-angular` does that 👍) but cloud foundry works as well!

PR #304, opened 2020-02-01
chore(examples): update dependencies

asudoh:
Also adds `.babelrc` to avoid transpiration of dependencies.


asudoh:
Thank you for reviewing @cal-smith! The Angular deploy failure seems of a timeout, but I saw a `node.js` heap error in local. Will investigate further.

asudoh:
The Angular deploy is fixed in: #321

PR #307, opened 2020-02-04
chore(devenv): fix knobs for angular/react

asudoh:
This change fixes Storybook knobs for Angular/React demo. It seems to have been caused by `start-storybook` command being Web Components version, where it used to be Vue version.

Using Angular/React versions of `start-storybook` does not seem to work somehow, so using Vue version for Angular/React for now.


asudoh:
Addressed by: #321

PR #309, opened 2020-02-06
chore(components): move by-state props to enums

asudoh:
This change replaes misc boolean attributes with enum attribute. This ensures forward-compatibility in case additional color scheme option may be added.

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -42,6 +42,21 @@ export enum BUTTON_KIND {
 }
 
 /**
+ * Button size.
+ */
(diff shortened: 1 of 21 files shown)


src/components/button/button.ts
```diff
@@ -130,10 +145,10 @@ class BXButton extends FocusMixin(LitElement) {
   rel!: string;
 
   /**
-   * `true` if the button should be a small variant.
+   * Butotn size.
```
emyarod:
```suggestion
   * Button size.
```

PR #321, opened 2020-02-19
chore(devenv): build fixes

asudoh:
This change ensures that Storybook build works well with the Storybook CLI commands for the corresponding framework.


asudoh:
Thank you for reviewing @cal-smith!

PR #338, opened 2020-03-07
chore(devenv): add more docs to Storybook

asudoh:
Adds `.mdx` to the Storybook environment so that `carbon-custom-elements` users can better understand how to use the library by looking at the Storybook environment.

image


cal-smith:
This is very exciting 😁

PR #359, opened 2020-04-10
chore(package): add type field

asudoh:
This will enable Node to import our modules directly without requiring a transpiled version, though we have few modules that is meant to be used in non-browser environment.


asudoh:
Oops I seem to have jumped the gun - It appears that need to update the dev/build env as well. Closing for now.

PR #378, opened 2020-05-01
Add min and max validity to number input

### LINKED ISSUE #375: Add out-of-range/invalid messages to BXNumberInput
Based on the work in: https://github.com/carbon-design-system/carbon-custom-elements/pull/374

cal-smith:
closes #375

Basically just follows the pattern we've been using everywhere as far as determining and showing the validity messages. This does add two new slots and properties for the min/max validity messages, as well as an additional enum just for the number input states.


src/components/number-input/number-input.scss
```diff
@@ -14,4 +14,8 @@ $css--plex: true !default;
   // the base text-input class has a 100% so we need to propagate it up to the host element
   width: 100%;
   outline: none;
+
+  .bx--form-requirement.bx--hidden {
```
asudoh, line 18: Wondered if we can use hidden attribute (`[hidden]` selector) here.

src/components/number-input/number-input.ts
```diff
@@ -229,6 +276,16 @@ export default class BXNumberInput extends BXInput {
             ${this.validityMessage}
```
asudoh, line 276: Should we hide this section if we should out-of-range validity message?

src/components/number-input/number-input.scss
cal-smith: I actually tried that first! For some reason it didn't hide slotted content ... I'm not sure if that's a pattern that's been successful elsewhere though

src/components/number-input/number-input.ts
cal-smith: I'm not sure ... I was debating that too ... It may make sense to show it, since it's still invalid, but since the range error is a specific type of invalid maybe it would be better to make it clearer that it's just the range that's an issue?

asudoh: Yeah, probably if the validity message is something like "The value should be less than 100" user knows that user did something wrong as well as how to fix it...?

src/components/number-input/number-input.ts
```diff
@@ -51,6 +59,27 @@ export default class BXNumberInput extends BXInput {
...
+      return NUMBER_INPUT_VALIDATION_STATUS.EXCEEDED_MAXIMUM;
+    }
+    if (this._input?.valueAsNumber < Number(this.min)) {
+      return NUMBER_INPUT_VALIDATION_STATUS.EXCEEDED_MINIMUM;
+    }
+    return super._testValidity();
+  }
+
+  _getValidityMessage(state: string) {
+    if (Object.keys(NUMBER_INPUT_VALIDATION_STATUS).includes(state)) {
```
asudoh, line 73:
```suggestion
    if (Object.values(NUMBER_INPUT_VALIDATION_STATUS).includes(state)) {
```

Seems that `Object.keys(NUMBER_INPUT_VALIDATION_STATUS)` returns `'EXCEEDED_MAXIMUM'` whereas the `state` is `'exceeded_maximum'`.

src/components/number-input/number-input.ts
```diff
@@ -229,6 +276,16 @@ export default class BXNumberInput extends BXInput {
             ${this.validityMessage}
           </slot>
         </div>
+        <div class="${validityMaxClasses}">
+          <slot name="validity-message-max">
+            ${this.validityMessageMin}
```
asudoh, line 281:
```suggestion
            ${this.validityMessageMax}
```

src/components/number-input/number-input.ts
```diff
@@ -229,6 +276,16 @@ export default class BXNumberInput extends BXInput {
...
+          </slot>
+        </div>
+        <div class="${validityMinClasses}">
+          <slot name="validity-message-min">
+            ${this.validityMessageMax}
```
asudoh, line 286:
```suggestion
            ${this.validityMessageMin}
```

src/components/number-input/number-input.scss
asudoh: Just a hunch... Any chance it was due to missing `?` prefix at `?hidden="${validity !== NUMBER_INPUT_VALIDATION_STATUS.EXCEEDED_MINIMUM}"`?

```suggestion
  .bx--form-requirement[hidden] {
```

I could make it work with the combination of above two somehow. I may be wrong, though.

src/components/number-input/number-input.ts
```diff
@@ -123,6 +152,12 @@ export default class BXNumberInput extends BXInput {
   @property({ attribute: 'decrement-button-assistive-text' })
   decrementButtonAssistiveText = 'decrease number input';
 
+  @property({ attribute: 'validity-message-max' })
```
asudoh, line 155: If possible, could you add comments to the two new attributes? Doing so makes the comment appear in the property table in Docs page in Storybook.

src/components/number-input/number-input.scss
cal-smith: Hmm ... definitely had the `?` 🤔 I didn't have the styles when I tried it, so that would definitely make it work

src/components/number-input/number-input.ts
cal-smith: Doh 🤦 yeah that would make sense

src/components/number-input/number-input.ts
cal-smith: For sure!

src/components/number-input/number-input.ts
cal-smith: Yeah ... So, maybe for now we'll exclude the standard message since that seems like a sensible default

src/components/number-input/number-input.ts
```diff
@@ -70,7 +70,7 @@ export default class BXNumberInput extends BXInput {
   }
 
   _getValidityMessage(state: string) {
-    if (Object.keys(NUMBER_INPUT_VALIDATION_STATUS).includes(state)) {
+    if (Object.values(NUMBER_INPUT_VALIDATION_STATUS).includes(state)) {
```
asudoh, line 73:
```suggestion
    if (Object.values(NUMBER_INPUT_VALIDATION_STATUS).includes(state as NUMBER_INPUT_VALIDATION_STATUS)) {
```

This may solve the TS check error... Sorry if it's not the case.

asudoh:
Thanks @cal-smith for the updates! 🎉

PR #379, opened 2020-05-01
feat(side-nav): introduce responsive mode

### LINKED ISSUE #377: UI shell: Collapsed side nav without rail
https://www.carbondesignsystem.com/components/UI-shell-header/code shows an example with side nav collapsed without its "rail" state. This issue is to ensure that use case can be demonstrated. May need some code, or may be done only with demonstration code.

### LINKED ISSUE #380: Side nav: One for representing header menu in narrower screen
One seen at: https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-navigation

asudoh:
This change introduces a mode in `<bx-side-nav>` to allow itself to be collapsed completely, instead of showing "rail" in its collapsed state.

Also introduces a custom event to `<bx-header-menu-button>` to tell its toggle state. It replaces the mechanism where `<bx-header-menu-button>` used to update `<bx-side-nav>` directly.

This change also introduces another mode in `<bx-side-nav>` to use it as a substitute of `<bx-header-nav>`, which is, `<bx-side-nav usage-mode="header-nav">`.

With `usage-mode="header-nav"`, side nav is shown when header nav is hidden, and side nav is hidden when header nav is shown.

Application using `usage-mode="header-nav"` should take care of mirroring `<bx-header-nav>` content to `<bx-side-nav>`.

Fixes #377.
Fixes #380.


abdonrd:
> Deploy preview for _carbon-custom-elements_ ready!
> 
> Built with commit [8733617](https://github.com/carbon-design-system/carbon-custom-elements/commit/87336175ceff3f67daf1747fcd74e20d19580182)
> 
> https://deploy-preview-379--carbon-custom-elements.netlify.app

@asudoh this demo doesn't work for me.

asudoh:
@abdonrd Thank you for taking a look - Seems that it's working for me. I think https://deploy-preview-379--carbon-custom-elements.netlify.app/?path=/story/components-ui-shell--header is the most representative example, where the collapse button appears when you make browser narrower. That said, the new (default) mode makes side nav collapsible only in narrower screen (as dictated by the style guide).

abdonrd:
I see this:

ezgif-4-cd7b696a4b0b

I want to have the nav menu on desktop and drawer menu on mobile, like this:
covid19-hpc-consortium.org

asudoh:
@abdonrd Would you try making your browser window narrower? I think the hamburger icon will appear at the top left and you can change the expanded state with a knob, too:

side-nav-responsive

The COVID site seems to be employing a similar responsive pattern, too.

abdonrd:
Got it! You're right!

abdonrd:
@asudoh And there is an option to have the drawer hidden in desktop?

asudoh:
I don't the design guide allows hiding the side nav in wider screen, but would you want to elaborate below a bit more?

> Because in this moment we already have the menu in the toolbar.

abdonrd:
Screenshot 2020-05-01 at 17 14 14

When we share the same nav menu in the toolbar and in the drawer...
We don't want to show the drawer in the desktop mode, because we already have the toolbar menu.

asudoh:
Is your requirement similar to this? https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-navigation

abdonrd:
> Is your requirement similar to this? https://react.carbondesignsystem.com/?path=/story/ui-shell--header-base-w-navigation

Yes!

asudoh:
OK given it's a slightly different requirement from what's implemented here, created: #380

asudoh:
Added a commit to fix #380, which is available with selecting "For header nav" for `usage-mode` knob.

abdonrd:
> Added a commit to fix #380, which is available with selecting "For header nav" for `usage-mode` knob.

Awesome, thanks @asudoh! :👏

abdonrd:
Why the new options doesn't appear here?

https://custom-elements.carbondesignsystem.com/?path=/story/components-ui-shell--header

asudoh:
@abdonrd Good question, having trouble with deploying the latest stuff to custom-elements.carbondesignsystem.com. Meanwhile we can use carbon-custom-elements.netlify.app.

PR #422, opened 2020-06-18
chore(dependencies): upgrade to Carbon 10.14

asudoh:


jeffchew:
@asudoh LGTM, though I think we should figure out what we want to do with automated visual regression. Something like percy would be great here, but this falls more on the Carbon jurisdiction and should pay for a testing plan there perhaps?

asudoh:
@jeffchew Good point, this repo basically relies on unit test and [UI integration test](https://github.com/carbon-design-system/carbon-custom-elements/tree/v1.0.0/tests/integration/ui) on this, but I'm open for introducing Percy to this repo especially if the license covers this repo.

jeffchew:
> @jeffchew Good point, this repo basically relies on unit test and [UI integration test](https://github.com/carbon-design-system/carbon-custom-elements/tree/v1.0.0/tests/integration/ui) on this, but I'm open for introducing Percy to this repo especially if the license covers this repo.

The license surely does, but the cost is based on number of snapshots taken. We are on a 200k snapshots per month plan which is covered by our react and now web components package, which I think will take up a large portion of that. I think Carbon core team is looking into percy as a solution, but I think is in the trial periods.

PR #430, opened 2020-07-06
feat(file-uploader): introduction

asudoh:


src/components/file-uploader/file-uploader-shell.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const { prefix } = settings;
+
+/**
+ * The shell UI for file uploader.
+ * @element bx-file-uploader-shell
+ * @slot helper-text The helper text.
+ * @slot label-text The label text.
+ */
+@customElement(`${prefix}-file-uploader-shell`)
```
cal-smith, line 22: I think we should be good to just have
```suggestion
@customElement(`${prefix}-file-uploader`)
```
for this - the rest of the components just slot into this one as needed anyway.

asudoh: @cal-smith Thank you for your thoughts - Makes sense to me, updated.

asudoh:
Merging as all review comments are addressed, but @cal-smith please don't hesitate to let me know if you have any further comments, and I'll address.

PR #439, opened 2020-07-09
chore(build): add CF deploy

asudoh:


abdonrd:
Could you publish the GitHub Action as separated repository?

It would be great! We could use it in other projects too!

asudoh:
@abdonrd My manager @jeffchew is positive on that, let me dig into this further.

asudoh:
@abdonrd Created https://github.com/carbon-design-system/action-ibmcloud-cf, please see the README there for the details. Updated this PR to use that action.

abdonrd:
Awesome, thanks @asudoh! 👏

PR #450, opened 2020-07-20
fix(ui-shell): default label text

asudoh:
This change adds default label text for humberger menu button for side nav.

Refs carbon-design-system/ibm-dotcom-library#3161.

(Requires #446 to pass the CI)


asudoh:
@jeffchew Yes, you are correct - Thank you for reviewing!

PR #457, opened 2020-07-31
fix(link): change disabled handling

asudoh:
Changes `<bx-link>`, `<bx-btn>` (with `href` attribute/property) to make sure following link with context menu is prevented when they are disabled.

Also adds `disabled` and some other attributes/properties to `<bx-clickable-tile>`.

Refs carbon-design-system/ibm-dotcom-library#3405.


asudoh:
Thanks @RaphaelAmadeu for reviewing!

PR #460, opened 2020-08-04
fix(bx-side-nav): focus on toggle for the first element

RaphaelAmadeu:


src/components/ui-shell/side-nav.ts
```diff
@@ -81,6 +81,9 @@ class BXSideNav extends HostListenerMixin(LitElement) {
    */
   private _handleButtonToggle(event: CustomEvent) {
     this.expanded = event.detail.active;
+    if (this.expanded) {
+      this.querySelectorAll((this.constructor as typeof BXSideNav).selectorNavItems)[0].focus();
```
asudoh, line 85:
```suggestion
      (this.querySelector((this.constructor as typeof BXSideNav).selectorNavItems) as HTMLElement).focus();
```

* `.querySelector(selector)` returns the first match, and thus equivalent to `.querySelectorAll(selector)[0]`.
* `.querySelector()` returns an `Element`, but `Element` doesn't have `.focus()`. `HTMLElement` (an inheritance of `Element`) does. (Meaning, doing so avoids TS error)

RaphaelAmadeu: Thanks for the insights!

src/components/ui-shell/side-nav.ts
```diff
@@ -81,6 +81,9 @@ class BXSideNav extends HostListenerMixin(LitElement) {
    */
   private _handleButtonToggle(event: CustomEvent) {
     this.expanded = event.detail.active;
+    if (this.expanded) {
+      (this.querySelector((this.constructor as typeof BXSideNav).selectorNavItems) as HTMLElement).focus();
```
asudoh, line 85: My apologies, forgot to cover a condition where the focusable element is not found:

```suggestion
      (this.querySelector((this.constructor as typeof BXSideNav).selectorNavItems) as HTMLElement)?.focus();
```

PR #461, opened 2020-08-04
chore(package): rename package

asudoh:
Upon the community's consemsus, changes `carbon-custom-elements` package name to `carbon-web-components`, to resonate to the umbrella term and to the communities around Web Components technologies.


asudoh:
Thanks @jeffchew for reviewing! You are correct about the remainder of the works, with an addition to deprecate the older package name.

PR #476, opened 2020-08-26
feat(timeout): new property for the notification component

### LINKED ISSUE #291: Notification proposal
Something that we are trying to achieve with the notification component is to close the notification after X ms but by now and from what I could investigate through the code and storybook the library doesn't have yet a `open` and a `close` event to modify the notification behaviour. 

Looking into the [spec](https://www.carbondesignsystem.com/components/notification/code) for this component it seems that it has a `remove` event to handle this kind of things. I don't know if it can be possible to implement it.

Tansito:
### Related Ticket(s)

Fix #291 

### Description

This changes add the property `timeout` to the components `inline-notification` and `toast-notification`. This new property will close the notification if a time (in ms) is specified after that time pass.

### Changelog

**New**

- New property `timeout` for: `inline-notification`, `toast-notification`.

No breaking changes.

### DIFF
--- a/src/components/notification/inline-notification.ts
+++ b/src/components/notification/inline-notification.ts
@@ -86,20 +86,12 @@ class BXInlineNotification extends FocusMixin(LitElement) {
   /**
    * Current timeout identifier
    */
-  protected _timeoutID?: number | null;
+  protected _timeoutID: number | null = null;
 
   /**
    * Notification type.
    */
   protected _type = NOTIFICATION_TYPE.INLINE;
-
-  /**
-   * Creates a new timeout
-   * @param timeout the time in ms
-   */
-  protected _setTimeout(timeout: number) {
-    return setTimeout(this._handleInitiatedCloseEvent.bind(this), timeout);
-  }
 
   /**
    * Cancels the current timeout configured for the notification
@@ -118,7 +110,27 @@ class BXInlineNotification extends FocusMixin(LitElement) {
     if (this._timeoutID) {
       this._cancelTimeout(this._timeoutID);
     }
-    this._timeoutID = this._setTimeout(timeout);
+    this._timeoutID = setTimeout(this._handleTimerInitiatedClose.bind(this), timeout);
+  }
+
+  /**
+   * Handles timeout close request of this modal.
+   */
+  protected _handleTimerInitiatedClose(triggeredBy: undefined) {
+    if (this.open) {
+      const init = {
+        bubbles: true,
+        cancelable: true,
+        composed: true,
+        detail: {
+          triggeredBy,
+        },
+      };
+      if (this.dispatchEvent(new CustomEvent((this.constructor as typeof BXInlineNotification).eventBeforeClose, init))) {
+        this.open = false;
+        this.dispatchEvent(new CustomEvent((this.constructor as typeof BXInlineNotification).eventClose, init));
+      }
+    }
   }
 
   /**
@@ -126,14 +138,14 @@ class BXInlineNotification extends FocusMixin(LitElement) {
    * @param event The event.
    */
   protected _handleClickCloseButton({ target }: MouseEvent) {
-    this._handleInitiatedCloseEvent(target);
+    this._handleUserInitiatedClose(target);
   }
 
   /**
-   * Handles initiated close request of this modal.
-   * @param triggeredBy The element that triggered this close request if there is one.
+   * Handles user-initiated close request of this modal.
+   * @param triggeredBy The element that triggered this close request.
    */
-  protected _handleInitiatedCloseEvent(triggeredBy?: EventTarget | null) {
+  protected _handleUserInitiatedClose(triggeredBy: EventTarget | null) {
     if (this.open) {
       const init = {
         bubbles: true,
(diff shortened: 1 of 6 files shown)


Tansito:
I have a little question on this PR @asudoh , in the next [line](https://github.com/carbon-design-system/carbon-web-components/pull/476/files#diff-cd776c2b74aff035d97cfc6bab6b0abcR84). Meanwhile this initial setup for the test it works I think it's not the best approach, probably.

For this use case I saw that Jest has [Timer mocks](https://jestjs.io/docs/en/timer-mocks) but I couldn't make them work properly I don't know if you have a little more experience working with them, Akira. Thanks in advance as always!

asudoh (commented):
Thanks @Tansito for jumping in! As I see the code, the following cases comes to my mind:

* The notification gets closed then re-opened during the timeout for auto-close is in flight
* The timeout value is changed during the timeout for auto-close is in flight

To cope with such kind of cases, I tend to keep the timeout handle (what `setTimeout()` returns) in a private property and clear the timeout when `open` or `timeout` changes.

I also tend to null-ify such private value upon clearing the timeout or when the timer fires.

For your question wrt timer, I tend to create a private method for the logic creating a timer, and mock such private method in the test.

Hope above helps!

src/components/notification/inline-notification.ts
```diff
@@ -83,11 +83,56 @@ const iconsForKinds = {
...
+    if (this._timeoutID) {
+      this._cancelTimeout(this._timeoutID);
+    }
+    this._timeoutID = setTimeout(this._handleTimerInitiatedClose.bind(this), timeout);
+  }
+
+  /**
+   * Handles timeout close request of this modal.
+   */
+  protected _handleTimerInitiatedClose(triggeredBy: undefined) {
```
asudoh, line 119: Wondering if you wanted to make `undefined` as the default value, instead of making it the type of `triggeredBy`? If so, we can:

```suggestion
  protected _handleTimerInitiatedClose(triggeredBy?: Element) {
```

src/components/notification/inline-notification.ts
```diff
@@ -195,6 +240,12 @@ class BXInlineNotification extends FocusMixin(LitElement) {
   @property({ type: Boolean, reflect: true })
   open = true;
 
+  /**
+   * Notification time in ms until gets closed.
+   */
+  @property({ type: Number, reflect: true })
+  timeout?: number | null;
```
asudoh, line 247:
```suggestion
  timeout: number | null = null;
```

src/components/notification/inline-notification.ts
Tansito: This was to clarify how should I handle the callbacks. With the last refactorization should be clearer.

src/components/notification/inline-notification.ts
```diff
@@ -83,24 +83,49 @@ const iconsForKinds = {
...
+
+  /**
+   * Overrides (if exists) the timeout to close the notification
+   * @param timeout the time in ms
+   */
+  protected _initializeTimeout(timeout: number) {
+    if (this._timeoutID) {
+      this._cancelTimeout(this._timeoutID);
+    }
+    this._timeoutID = setTimeout(this._handleUserOrTimerInitiatedClose.bind(this), timeout);
```
asudoh, line 113:
```suggestion
    this._timeoutID = setTimeout(this._handleUserOrTimerInitiatedClose.bind(this, null), timeout);
```

src/components/notification/inline-notification.ts
```diff
@@ -83,24 +83,49 @@ const iconsForKinds = {
...
   }
 
   /**
-   * Handles user-initiated close request of this modal.
-   * @param triggeredBy The element that triggered this close request.
+   * Handles user-initiated or through timer close request of this modal.
+   * @param triggeredBy The element that triggered this close request, if there is one.
    */
-  protected _handleUserInitiatedClose(triggeredBy: EventTarget | null) {
+  protected _handleUserOrTimerInitiatedClose(triggeredBy?: EventTarget | null) {
```
asudoh, line 128:
```suggestion
  protected _handleUserOrTimerInitiatedClose(triggeredBy: EventTarget | null) {
```

asudoh (commented):
Almost there - I now see why we had the last argument list for `_handleUserOrTimerInitiatedClose()`. Wondering the new suggestion makes sense. Thanks @Tansito!

src/components/notification/inline-notification.ts
Tansito: The problem that I see with this is that when the `setTimeout` is launched the value in `triggeredBy` is **undefined** so I think is interesting to specify that the property can be **undefined**. How do you usually handle that?

Tansito: Ok, nevermind. I just saw your last proposal!

asudoh (approved):
LGTM 👍 - Thank you @Tansito for all the updates!

src/components/notification/inline-notification.ts
```diff
@@ -83,24 +83,49 @@ const iconsForKinds = {
...
+
+  /**
+   * Overrides (if exists) the timeout to close the notification
+   * @param timeout the time in ms
+   */
+  protected _initializeTimeout(timeout: number) {
+    if (this._timeoutID) {
+      this._cancelTimeout(this._timeoutID);
+    }
+    this._timeoutID = setTimeout(this._handleUserOrTimerInitiatedClose.bind(this, null), timeout);
```
asudoh, line 113:
```suggestion
    this._timeoutID = (setTimeout(this._handleUserOrTimerInitiatedClose.bind(this, null), timeout) as unknown) as number;
```

Issue #229, opened 2019-12-13
Component: Select


asudoh:

RobertaJHahn:
@asudoh Why is this issue in Review?   It doesn't contain a description and is not assigned to this sprint.

asudoh:
@RobertaJHahn Oh probably I misunderstood the purpose of the pipeline; I put it simply because there is PR in review.

RobertaJHahn:
@asudoh You are correct about the purpose of the pipeline, but as you have observed in the meetings we use labels and assignments to track the work and do reporting.  If you are working on this issue, then you should assign it to yourself.  If you are working on it in the current sprint, then the sprint should be assigned.  We also use labels to categorize the work between design and dev, web components and react.  Is this a React or Web Components component?  Could you add this label?  `package: react` or `package: web components`

I also add the issue to an Airtable where the velocity is tracked.  

Let me know if you have any questions.

asudoh:
Thanks for the info @RobertaJHahn! Added the label. This is an old issue that was created before `carbon-web-components` repo is integrated into our project scheme.

RobertaJHahn:
@asudoh I noticed that after I posted the comment about the component label.  When moving an issue from one repo to another you loose all of the information associated with the issue.  (milestone, labels, release....)  I have gotten burned by that situation in the past.  Thanks for the update.

asudoh:
Work finished, the feature available at: https://carbon-web-components.netlify.app

PR #478, opened 2020-08-29
feat(select): introduction

asudoh:
### Related Ticket(s)

Refs #229.

### Description

Introduces `<bx-select>` that works as a Carbon-styled `<select>`.

This works with `<bx-select-item>` and `<bx-select-item-group>`. When they are put in `<bx-select>`, `<bx-select>` creates the corresponding `<option>` and `<optgroup>`, respectively, into shadow DOM. Doing so allows the shadow DOM style of `<bx-select>` to control the style of the `<option>` and `<optgroup>`, notably the disabled ones.

Also introduces Angular directive for `<bx-select>`.

### Changelog

**New**

- `<bx-select>` that works as a Carbon-styled `<select>`.
- `<bx-select-item>`/`<bx-select-item-group>`, a pseudo custom element that lets `<bx-select>` create corresponding `<option>`/`<optgroup>`, respectively.
- An Angular directive for `<bx-select>`.


jeffchew (approved):
Sorry it took awhile to get to this @asudoh !

Looks good to me! There are some minor thoughts though when I was doing some quick comparisons to the React version:
- There's no default value when first loading in the component, React has the value `Choose an option`. Is that configurable for this one?
- The `validity-message` is blank by default in storybook, maybe have it set as `A valid value is required` to match React by default?

asudoh:
@jeffchew Thank you for reviewing! Fixed the placeholder. Wrt validity message, storybook in our codebase doesn't provide the default value across the component, but we can consider a change across components if we have a strong desire for it.

PR #480, opened 2020-09-04
chore(dependencies): upgrade Carbon

asudoh:
### Description

To `10.19.x`.

### Changelog

**Changed**

- Upgrades Carbon to `v10.19.x`.


jeffchew (approved):
@asudoh offhand this looks fine to me, but we need to get Percy running on here. The only thing I'm worried about at the moment is more about cost, as core carbon team is also temporarily piggybacking on our account and also increasing the snapshot usage.

jeffchew:
@asudoh FYI, one thing we picked up in the visual regression of Carbon expressive is that tabs now behave differently for smaller window sizes. It no longer breaks into a dropdown:

https://percy.io/538fc19a/Carbon-Expressive-Theme/builds/6774657

https://react.carbondesignsystem.com/?path=/story/tabs--default

asudoh:
@jeffchew Thank you for reviewing - So, with your concern on cost, which way would you prefer for now?

1. Biting the bullet on the cost and add Percy here
2. Keep doing manual testing (Like what I did for this PR)

Wrt React your finding is correct, and it's the change Carbon core introduced. They replaced dropdown with horizontally scrollable tab (with scrolling buttons as far as I can remember).

PR #483, opened 2020-09-15
fix(side-nav): focus on end-animation

asudoh:
### Related Ticket(s)

Refs carbon-design-system/ibm-dotcom-library#3940.

### Description

This change ensures a side nav item is ready to be focused on before side nav sets focus on the first nav item when it's expanded.

### Changelog

**New**

- A mechanism to wait for `transitionend` event before `<bx-side-nav>` sets focus on the first nav item when it's expaneded.


jeffchew (approved):
LGTM!

PR #484, opened 2020-09-17
fix(build): externalize Babel helpers

asudoh:
### Description

This change fixes an issue where Babel helpers are inlined in each of our modules where they should have been ESM `import`s.

Also adds a build step to minify `lit-html` templates.

### Changelog

**New**

- A build step to minify `lit-html` templates.

**Changed**

- Fix for an issue where Babel helpers are inlined in each of our modules where they should have been ESM `import`s.


kennylam (approved):
Looks good thanks @asudoh 👍

PR #502, opened 2020-09-26
chore(dropdown-item): add shadow parts

asudoh:
### Changelog

**New**

- Shadow parts to `<bx-dropdown-item>`.


kennylam (approved):
Thanks, looks good @asudoh! Quick question, are `carbon-web-components` meant to align with the expressive theme? Because I believe icons should be `20px`.

asudoh:
Thank you for reviewing @kennylam!

> Quick question, are `carbon-web-components` meant to align with the expressive theme? Because I believe icons should be `20px`.

`carbon-web-components` is based on Carbon core design and thus the base icon size should be same as core Carbon (`16px`). In future, good to support changing base icon size via `--cds-icon-size-01` CSS custom property.

kennylam:
> `carbon-web-components` is based on Carbon core design and thus the base icon size should be same as core Carbon (`16px`). In future, good to support changing base icon size via `--cds-icon-size-01` CSS custom property.

Got it, thanks @asudoh!

PR #521, opened 2020-10-02
chore(format): introduce eslint-plugin-lit

asudoh:
### Description

Introduce `eslint-plugin-lit` so we can catch invalid `lit-html` template syntax, notably usage of self-closing tags for non-void elements.

### Changelog

**New**

- `eslint-plugin-lit` and its configuration.


src/.eslintrc
```diff
@@ -3,7 +3,11 @@
     "ecmaVersion": 6,
     "sourceType": "module"
   },
+  "plugins": [
+    "lit"
+  ],
   "extends": [
```
abdonrd, line 9: With the `"plugin:lit/recommended"` you don't need to add the plugin:

https://github.com/43081j/eslint-plugin-lit/blob/master/src/configs/recommended.ts#L2

asudoh: Oh thanks @abdonrd! Updated.

abdonrd (approved):

PR #522, opened 2020-10-02
fix(modal): better focus-wrapping

asudoh:
### Related Ticket(s)

Refs carbon-design-system/ibm-dotcom-library#4057.

### Description

This change fixes an issue found in `@carbon/ibmdotcom-web-components` library, where `<bx-modal>`'s focus-wrapping code attempts to focus on a focusable custom element but cannot detect it as focusable because the host element does not have a box. Replaces a logic using box size for "focusable" detection with "try-to-focus" logic.

Also:

* Fixes an issue with browsers that don't support `delegatesFocus` where the host element as well as an element in shadow DOM gets keyboard focus as user hits tab key within modal.
* Fixes a race condigion with browsers that don't support `delegatesFocus` where attempts to focus on a custom element that has a focusable element in it fails, seen with a `BXModal` inheritance.

### Changelog

**Changed**

- Replaces a logic using box size for "focusable" detection with "try-to-focus" logic.
- Fixes an issue with browsers that don't support `delegatesFocus` where the host element as well as an element in shadow DOM gets keyboard focus as user hits tab key within modal.
- Fixes a race condigion with browsers that don't support `delegatesFocus` where attempts to focus on a custom element that has a focusable element in it fails


kennylam (approved):
Looks good thanks @asudoh!

Issue #538, opened 2020-11-03
Move deploy previews to IBM Cloud


jeffchew:
#### User Story

> As a `[user role below]`:
Pull request reviewer

> I need to:
see a deploy preview of the pull request code

> so that I can:
test the code while reviewing without any quota limitations

#### Additional information

- Carbon Web Components currently uses Netlify, which is on the OSS tier and has very limited build time per month
- Existing script used for creating deploy preview comments in Carbon for IBM.com: https://github.com/carbon-design-system/carbon-for-ibm-dotcom/blob/master/tasks/deploy_preview_comment.js

#### Acceptance criteria

- [x] Move deploy preview creation to IBM Cloud
- [x] Should allow for previews created from forked repos
- [ ] Deactivate Netlify

jeffchew:
Ended up taking this one in as I couldn't complete out https://github.com/carbon-design-system/carbon-web-components/issues/472 without it.

Issue #472, opened 2020-08-20
Set up Percy


asudoh:
### The problem

It's hard to sometimes prevent a visual regression caused by change in code.

### The solution

Introduce automated visual regression too, like Percy.

PR #541, opened 2020-11-13
chore(percy): adding percy automated visual testing

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-web-components/issues/472, https://github.com/carbon-design-system/carbon-web-components/issues/538

### Description

This PR adds Percy/Browserstack automated viisual regression testing.
The essential Percy Storybook artifact is added, as well as the Github
Action to update the baseline whenever master or release branches are
updated.

The `inPercy` method has been added, though currently not in use
anywhere yet. This will be useful in case any percy-only conditions need
 to be made in order to avoid false positives in percy snapshots.

In addition, deploy previews have been added to the Carbon for IBM.com's CI solution, so Netlify can be removed from the workflow (@asudoh not sure if you're the only one that has access to that account?)

### Changelog

**New**

- Percy snapshot testing (deploy previews and basline updating)

**Changed**

- Updated `header-max-length` for commitlint settings to 72 characters


README.md
```diff
@@ -21,6 +21,11 @@ A Carbon Design System variant that's as easy to use as native HTML elements, wi
     <img src="https://www.netlify.com/img/global/badges/netlify-color-accent.svg"/>
   </a>
 </p>
+<p align="center">
+    <a href="https://percy.io/538fc19a/Carbon-Web-Components">
```
asudoh, line 25: Just a little indent nit 😉:

```suggestion
  <a href="https://percy.io/538fc19a/Carbon-Web-Components">
```

asudoh (approved):
LGTM 👍 - Thanks @jeffchew for doing this!

jeffchew:
@asudoh I set up the standard deploy preview, but now just noticed that this kicked off as well:

Screen Shot 2020-11-12 at 8 43 16 PM

Will I need to set up those for deploy previews too?

asudoh:
@jeffchew Yes we create deploy previews for React, Angular and Vue in addition to vanilla. Thank you for asking!

jeffchew:
> @jeffchew Yes we create deploy previews for React, Angular and Vue in addition to vanilla. Thank you for asking!

Thanks! I don't see Netlify publishing a comment link though, but looks like it's getting created from the details link under checks. I'll see if I can add those during my day tomorrow.

I also noticed that there's some false positives getting generated from Percy. I am thinking I'll tell Percy to skip the skeleton stories as they have an animation that's triggering diffs (also will hit this up tomorrow).

Issue #526, opened 2020-10-09
Button: Add XL size variant


asudoh:
### The problem

Carbon core is adding XL button size variant.

### The solution

Implement it here.

PR #547, opened 2020-11-27
feat(button): add XL size variant

kennylam:
### Related Ticket(s)

#526 

### Description

Adds extra large size variant

### Changelog

**New**

- `xl` size variant

### DIFF
--- a/src/components/button/button-story.ts
+++ b/src/components/button/button-story.ts
@@ -33,7 +33,7 @@ const kinds = {
 const sizes = {
   'Regular size': null,
   [`Small size (${BUTTON_SIZE.SMALL})`]: BUTTON_SIZE.SMALL,
-  [`XL size (${BUTTON_SIZE.XLARGE})`]: BUTTON_SIZE.XLARGE,
+  [`XL size (${BUTTON_SIZE.EXTRA_LARGE})`]: BUTTON_SIZE.EXTRA_LARGE,
   [`Size for form field (${BUTTON_SIZE.FIELD})`]: BUTTON_SIZE.FIELD,
 };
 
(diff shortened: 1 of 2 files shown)


src/components/button/button.ts
```diff
@@ -70,6 +70,11 @@ export enum BUTTON_SIZE {
    */
   SMALL = 'sm',
 
+  /**
+   * X-Large size.
+   */
+  XLARGE = 'xl',
```
asudoh, line 76: To align to: https://github.com/carbon-design-system/carbon-web-components/blob/v1.8.0/src/components/dropdown/dropdown.ts#L82and https://github.com/carbon-design-system/carbon-web-components/blob/v1.8.0/src/components/input/input.ts#L46

```suggestion
  EXTRA_LARGE = 'xl',
```

src/components/button/button-story.ts
```diff
@@ -33,6 +33,7 @@ const kinds = {
 const sizes = {
   'Regular size': null,
   [`Small size (${BUTTON_SIZE.SMALL})`]: BUTTON_SIZE.SMALL,
+  [`XL size (${BUTTON_SIZE.XLARGE})`]: BUTTON_SIZE.XLARGE,
```
asudoh, line 36:
```suggestion
  [`XL size (${BUTTON_SIZE.XLARGE})`]: BUTTON_SIZE.EXTRA_LARGE,
```

asudoh (commented):
Thank you for jumping in @kennylam!

src/components/button/button.ts
kennylam: Thanks @asudoh!

src/components/button/button-story.ts
kennylam: Thanks @asudoh!

asudoh (approved):
Thanks @kennylam for your quick update!

Issue #548, opened 2020-12-01
Incorrect rendering of <bx-header> and <bx-side-nav> combination. Missing global styles?


pschyska:
## Detailed description

I'm experimenting with the UI shell components at the moment and I noticed some differences in rendering to the "official" components (e.g. react, Vanilla variants). I'm not sure if I'm doing something wrong.

My HTML is minimal:

````html
<!DOCTYPE html>
<html lang="en">
  <head>
    <title>m14r</title>
    <meta charset="utf-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <style>
      m14r-main {
        transition: opacity 200ms ease-out;
      }
      /* prevent FOUC */
      m14r-main:not(:defined) {
        opacity: 0;
      }
    </style>
  </head>
  <body>
    <m14r-main></m14r-main>
    <script type="module" src="./index.js"></script>
  </body>
</html>
````

My Component looks like this:

````ts
import {
  css,
  customElement,
  html,
  LitElement,
  TemplateResult,
} from "lit-element";
import "carbon-web-components/es/components/ui-shell/header";
import "carbon-web-components/es/components/ui-shell/header-name";
import "carbon-web-components/es/components/ui-shell/header-menu-button";
import "carbon-web-components/es/components/ui-shell/side-nav";
import "carbon-web-components/es/components/ui-shell/side-nav-link";

@customElement("m14r-main")
export class Main extends LitElement {
  static styles = css`
    * {
      box-sizing: border-box;
    }

    bx-header ~ bx-side-nav {
      margin-top: 3rem;
      height: calc(100% - 3rem);
    }

    bx-header ~ .bx--content {
      margin-left: 16rem;
      margin-top: 3rem;
      height: calc(100% - 3rem);
      width: calc(100% - 16rem);
    }
  `;

  render(): TemplateResult {
    return html`
      <bx-header>
        <bx-header-menu-button></bx-header-menu-button>
        <bx-header-name prefix="">M14R</bx-header-name>
      </bx-header>
      <bx-side-nav>
        <bx-side-nav-link href="javascript:void(0)" active
          >Item</bx-side-nav-link
        >
        <bx-side-nav-link href="javascript:void(0)"
          >Another Item</bx-side-nav-link
        >
      </bx-side-nav>
      <main class="bx--content">
        <h1>UI Shell Content</h1>

        <p>
          <strong>Pellentesque habitant morbi tristique</strong> senectus et
[...]

PR #549, opened 2020-12-02
docs(ui-shell): update doc of side nav

asudoh:
### Related Ticket(s)

Refs #548.

### Description

Adds a documentation on application-level style that should be added when `<bx-side-nav>` is used along with `<bx-header>`.

### Changelog

**New**

- A documentation on application-level style that should be added when `<bx-side-nav>` is used along with `<bx-header>`.


kennylam (approved):
Looks good thanks @asudoh!

PR #550, opened 2020-12-02
chore(components): introduce CommonJS version of React wrappers

asudoh:
### Related Ticket(s)

Refs carbon-design-system/carbon-for-ibm-dotcom#4251.

### Description

Introduces CommonJS version of React wrappers, that has the following characteristics:

1. Leverages "conditional exports" node.js (`package.json`) feature so `import 'carbon-web-components/es/components-react/tag/tag.js'` loads the corresponding CommonJS version in node.js environment (`/lib/components-react-node/tag/tag.js`). Ref: https://github.com/jkrems/proposal-pkg-exports#2-conditional-mapping
2. Does not load custom elements implemetation because custom elements implementation cannot be run in node.js environment.
3. Does not load `lit-html` or `lit-element` because they only ship with ECMAScript modules format.

As part of the effort, moves enums, etc. code to a separate modules so they can be built separately from React wrappers and we can make the logic of auto-generating CommonJS version of React wrappers simpler.

Also adds CodeSandbox example and their corresponding build integration tests for:

* Basic React SSR
* Next.js

### Changelog

**New**

- CommonJS version of React wrappers
- CodeSandbox example and their corresponding build integration tests for:
  - Basic React SSR
  - Next.js

**Changed**

- Moved enums, etc. in components to separate modules


kennylam (approved):
LGTM great change @asudoh  thanks!

PR #553, opened 2020-12-04
fix(react): fix serializers

asudoh:
### Description

Fixes a bug in number/object serializers in our React wrapper generator where sending `null` yielded to `'null'` in string, where it should have been kept in `null` in object.

### Changelog

**Changed**

- Fixes a bug in number/object serializers in our React wrapper generator where sending `null` yielded to `'null'` in string, where it should have been kept in `null` in object.


kennylam (approved):
LGTM thanks @asudoh!

PR #555, opened 2020-12-07
test(integration-ui): fix for WebKit

asudoh:
### Description

WebKit in the version of Playwright we are with seems to have a problem where `:host(bx-modal[open])` is defined before `:host(bx-modal)`.

This fixes the UI integration test failure caused by such problem.

### Changelog

**Changed**

- Workaround a problem in WebKit used in the Playwright version we are using that happens when 
`:host(bx-modal[open])` is defined before `:host(bx-modal)`.

### DIFF
--- a/src/components/modal/modal-story.scss
+++ b/src/components/modal/modal-story.scss
@@ -1,0 +1,14 @@
+//
+// Copyright IBM Corp. 2020
+//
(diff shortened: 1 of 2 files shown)


kennylam (approved):
LGTM thanks @asudoh!

Issue #533, opened 2020-10-22
Overflow menu: Add size variant


asudoh:
### The problem

Carbon core is adding size variant to overflow menu.

### The solution

Implement it here.

PR #560, opened 2020-12-10
feat(overflow-menu): add size variants

kennylam:
### Related Ticket(s)

#533 

### Description

This adds size variants to `overflow-menu`.

### Changelog

**New**

- `sm` and `xl` sizes

### DIFF
--- a/src/components/overflow-menu/overflow-menu.scss
+++ b/src/components/overflow-menu/overflow-menu.scss
@@ -13,6 +13,16 @@ $css--plex: true !default;
   @extend .#{$prefix}--overflow-menu;
 }
 
+:host(#{$prefix}-overflow-menu)[size='sm'] {
+  width: rem(32px);
+  height: rem(32px);
+}
+
+:host(#{$prefix}-overflow-menu)[size='xl'] {
+  width: rem(48px);
+  height: rem(48px);
+}
+
 :host(#{$prefix}-overflow-menu[disabled]) {
   cursor: not-allowed;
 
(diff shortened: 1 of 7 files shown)


src/components/overflow-menu/overflow-menu.scss
```diff
@@ -13,6 +13,16 @@ $css--plex: true !default;
   @extend .#{$prefix}--overflow-menu;
 }
 
+:host(#{$prefix}-overflow-menu)[size='sm'] {
+  width: rem(32px);
+  height: rem(32px);
```
asudoh, line 18:
```suggestion
  @extend .#{$prefix}--overflow-menu--sm;
```

src/components/overflow-menu/overflow-menu.scss
```diff
@@ -13,6 +13,16 @@ $css--plex: true !default;
...
+}
+
+:host(#{$prefix}-overflow-menu)[size='xl'] {
+  width: rem(48px);
+  height: rem(48px);
```
asudoh, line 23:
```suggestion
  @extend .#{$prefix}--overflow-menu--xl;
```

asudoh (commented):
Thank you for jumping in @kennylam! Please let me know the change suggestion doesn't work.

src/components/overflow-menu/overflow-menu.scss
kennylam: @asudoh Good call! Interesting that this works, in the past I've had problems extending classes directly from other imports.

asudoh (approved):
Thanks @kennylam for the update!

Issue #528, opened 2020-10-10
Content switcher: Add size variant


asudoh:
### The problem

Carbon core is adding size variant to content switcher.

### The solution

Implement it here.

PR #561, opened 2020-12-10
feat(content-switcher): add size variants

kennylam:
### Related Ticket(s)

#528  

### Description

This adds size variants to `content-switcher`.

### Changelog

**New**

- `sm` and `xl` sizes

### DIFF
--- a/src/components/content-switcher/content-switcher.scss
+++ b/src/components/content-switcher/content-switcher.scss
@@ -1,5 +1,5 @@
 //
-// Copyright IBM Corp. 2019
+// Copyright IBM Corp. 2019, 2020
 //
 // This source code is licensed under the Apache-2.0 license found in the
 // LICENSE file in the root directory of this source tree.
@@ -11,6 +11,14 @@ $css--plex: true !default;
 
 :host(#{$prefix}-content-switcher) {
   @extend .#{$prefix}--content-switcher;
+}
+
+:host(#{$prefix}-content-switcher[size='sm']) {
+  height: rem(32px);
+}
+
+:host(#{$prefix}-content-switcher[size='xl']) {
+  height: rem(48px);
 }
 
 :host(#{$prefix}-content-switcher-item) {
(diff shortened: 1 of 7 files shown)


src/components/content-switcher/content-switcher.scss
```diff
@@ -13,6 +13,14 @@ $css--plex: true !default;
   @extend .#{$prefix}--content-switcher;
 }
 
+:host(#{$prefix}-content-switcher[size='sm']) {
+  height: rem(32px);
```
asudoh, line 17:
```suggestion
  @extend .#{$prefix}--content-switcher--sm;
```

src/components/content-switcher/content-switcher.scss
```diff
@@ -13,6 +13,14 @@ $css--plex: true !default;
...
+}
+
+:host(#{$prefix}-content-switcher[size='xl']) {
+  height: rem(48px);
```
asudoh, line 21:
```suggestion
  @extend .#{$prefix}--content-switcher--xl;
```

asudoh (commented):
Thank you for jumping in @kennylam! Please let me know the change suggestion doesn't work.

src/components/content-switcher/content-switcher.scss
kennylam: @asudoh Good call! Interesting that this works, in the past I've had problems extending classes directly from other imports.

src/components/content-switcher/content-switcher.scss
kennylam: @asudoh Good call! Interesting that this works, in the past I've had problems extending classes directly from other imports.

asudoh: @kennylam Yeah it has to be a simple selector I believe but where it works we can avoid CSS rule duplicates.

asudoh (approved):
Thank you @kennylam for the update!

PR #563, opened 2020-12-16
fix(unordered-list): workaround for WebKit

asudoh:
### Description

This change works around a limitation in WebKit where `::before` pseudo element is not supported on `::slotted()` pseudo element.

Refs https://bugs.webkit.org/show_bug.cgi?id=178237.
Refs carbon-design-system/carbon-for-ibm-dotcom#4733.

### Changelog

**New**

- A workaround for a limitation in WebKit where `::before` pseudo element is not supported on `::slotted()` pseudo element.

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -29,6 +29,14 @@ $css--plex: true !default;
   }
 }
 
+:host(#{$prefix}-unordered-list) {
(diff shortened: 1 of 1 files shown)


kennylam (approved):
Looks good thanks @asudoh!

PR #573, opened 2021-01-13
fix(select): set option text so it is visible in mobile

annawen1:
### Related Ticket(s)

{{Provide url(s) to the related ticket(s) that this pull request addresses}}

### Description

Need to pass in the Option text as a child in order for it to be viewed in the native select on mobile

BEFORE:
Screen Shot 2021-01-13 at 5 48 56 PM

NOW:
Screen Shot 2021-01-13 at 5 49 33 PM

### Changelog

**New**

- {{new thing}}

**Changed**

- {{changed thing}}

**Removed**

- {{removed thing}}

### DIFF
--- a/src/components/select/select.ts
+++ b/src/components/select/select.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2020
+ * Copyright IBM Corp. 2020, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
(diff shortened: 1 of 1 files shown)


src/components/select/select.ts
```diff
@@ -97,6 +97,7 @@ class BXSelect extends ValidityMixin(FormMixin(LitElement)) {
                   ?selected="${selected}"
                   value="${ifNonNull(value)}"
                 >
+                  ${ifNonNull(textContent)}
```
asudoh, line 100:
```suggestion
                  ${textContent}
```

asudoh (commented):
LGTM, just one change suggestion - `ifNonNull` is only for attribute/properties. Thanks @annawen1 for the fix!

src/components/select/select.ts
annawen1: gotcha, committed that!

asudoh (approved):
Thanks @annawen1 for the update!

PR #574, opened 2021-01-19
chore(ci): update deploy script

asudoh:
### Description

Updates deploy script so that the GH pages doesn't use Jekyll.

### Changelog

**Changed**

- Deploy script so that the GH pages doesn't use Jekyll.


jeffchew (approved):
LGTM!

Issue #576, opened 2021-01-25
Number Input 'value' type error


Tansito:
## Detailed description

Using the component `number-input` when we run the typescript compiler with `tsc --noEmit` we obtain the next error from the console:

```
node_modules/carbon-web-components/es/components/number-input/number-input.d.ts:52:9 - error TS2611: 'value' is defined as a property in class 'BXInput', but is overridden here in 'BXNumberInput' as an accessor.

52     get value(): string;
```

> What version of Carbon Web Components are you using?

- Tested with the v1.11.1

## Steps to reproduce the issue

1. Just import the component:
```
import 'carbon-web-components/es/components/number-input/number-input';
```
2. Run the TS compiler: `tsc --noEmit`

## Additional information

- Screenshots or code
<img width="545" alt="Screen Shot 2021-01-25 at 6 38 30 PM" src="https://user-images.githubusercontent.com/9059044/105743125-a07aa200-5f3c-11eb-9b45-8f3588b9a01b.png">

asudoh:
The culprit is https://github.com/carbon-design-system/carbon-web-components/blob/v1.11.0/src/components/number-input/number-input.ts#L121, where it's not propagated `.d.ts` file and TypeScript folks say that it's an intended behavior. Need to see how to get around this.

PR #578, opened 2021-01-27
fix(number-input): fix TS error

asudoh:
### Related Ticket(s)

Refs #576.

### Description

Makes `.value` property of `<bx-input>` an accessor so `<bx-number-input>`, that inherits `<bx-input>`, doesn't have to redefine the type.

### Changelog

**Changed**

- A change to make `.value` property of `<bx-input>` an accessor so `<bx-number-input>`, that inherits `<bx-input>`, doesn't have to redefine the type.


annawen1 (approved):
LGTM!

jeffchew:
@asudoh looks like there is a merge conflict here, can you take a look?

asudoh:
Thanks @annawen1 and @jeffchew for reviewing - Fixed the merge conflict.

Issue #467, opened 2020-08-08
Data table: Tabbing away from batch action bar hides the bar


asudoh:
## Detailed description

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug.

> Is this issue related to a specific component?

Yes, data table.

> What browser are you working in?

Chrome `84.0.4147.105`

> What version of Carbon Web Components are you using?

`v1.4.0`

## Steps to reproduce the issue

1. https://web-components.carbondesignsystem.com/iframe.html?id=components-data-table--sortable&knob-Supports%20selection%20feature%20(has-selection)=true
2. Select a row
3. Click on a blank area above the table
4. Keep hitting tab key until the focus goes into the batch action bar
5. Keep hitting tab key until the focus goes out of the batch action bar

BUG: The batch action bar is hidden. Expected to see the batch action bar kept.

RobertaJHahn:
Added to Aug 31 Engineering meeting agenda.

RobertaJHahn:
@asudoh Just cleaning up the issues related to the Carbon Web Components January 13 release. This issue is included in the release, but doesn't appear to be finished. Should I move this issue to the next release in March?

cc: @jeffchew

RobertaJHahn:
@asudoh I just found my notes about the scheduling of this issue.  Moving to the March release.  

cc:  @jeffchew

PR #581, opened 2021-02-10
fix(data-table): corrected tabbing into proper bars

IgnacioBecerra:
### Related Ticket(s)
#467

### Description

Tabbing now focuses on the action bar that is currently displayed. Before the change, continuous tabbing kept focusing on the bar that was hidden, which effectively 'hid' the proper bar from view.

### Changelog

**Changed**

- `tabIndex` toggles on each action bar depending on which one is being shown

### DIFF
--- a/src/components/data-table/table-batch-actions.ts
+++ b/src/components/data-table/table-batch-actions.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2020
+ * Copyright IBM Corp. 2019, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -46,6 +46,12 @@ class BXTableBatchActions extends LitElement {
    */
   @property({ type: Number, attribute: 'selected-rows-count' })
   selectedRowsCount = 0;
+
+  updated(changedProperties) {
+    if (changedProperties.has('active')) {
+      this.setAttribute('tabindex', `${this.active ? 0 : -1}`);
+    }
+  }
 
   render() {
     const { formatSelectedItemsCount, selectedRowsCount, _handleCancel: handleCancel } = this;
--- a/src/components/data-table/table-batch-actions.ts
+++ b/src/components/data-table/table-batch-actions.ts
@@ -49,7 +49,7 @@ class BXTableBatchActions extends LitElement {
 
   updated(changedProperties) {
     if (changedProperties.has('active')) {
-      this.setAttribute('tabindex', `${this.active ? 0 : -1}`);
+      this.setAttribute('tabindex', `${this.active ? '' : '-1'}`);
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/components/data-table/table-batch-actions.ts
```diff
@@ -47,6 +47,12 @@ class BXTableBatchActions extends LitElement {
   @property({ type: Number, attribute: 'selected-rows-count' })
   selectedRowsCount = 0;
 
+  updated(changedProperties) {
+    if (changedProperties.has('active')) {
+      this.setAttribute('tabindex', `${this.active ? 0 : -1}`);
+    }
+  }
```
asudoh, line 54: Would you apply this roving `tabIndex` pattern to the button in Shadow DOM instead of the host element?

asudoh (commented):
Thank you for jumping in @IgnacioBecerra!

src/components/data-table/table-batch-actions.ts
IgnacioBecerra: Hm, I tried applying it to the button and any other divs that could be focused, but it seems it'd have to be applied to the host in order for it to work. It seems that each of the slotted `bx-btn` load without the default `tabindex`, making them able to be selected even if all the other elements have `tabindex="-1"`

Just as an FYI, the reason for this change is that in the default Sortable storybook variation, tabbing into the component makes the current bar go away, just like in the original issue that was opened, except with the opposite bar.

src/components/data-table/table-batch-actions.ts
```diff
@@ -47,6 +47,12 @@ class BXTableBatchActions extends LitElement {
   @property({ type: Number, attribute: 'selected-rows-count' })
   selectedRowsCount = 0;
 
+  updated(changedProperties) {
+    if (changedProperties.has('active')) {
+      this.setAttribute('tabindex', `${this.active ? 0 : -1}`);
```
asudoh, line 52: The reason why I was against adding `tabindex` to the host element is because we don't want the host element itself to be focused - We want one of the buttons, etc. in the Shadow DOM to be focused instead. Seems that making the attribute value empty works for that purpose.

```suggestion
      this.setAttribute('tabindex', `${this.active ? '' : '-1'}`);
```

src/components/data-table/table-toolbar-content.ts
```diff
@@ -25,6 +25,12 @@ class BXTableToolbarContent extends LitElement {
   @property({ type: Boolean, reflect: true, attribute: 'has-batch-actions' })
   hasBatchActions = false;
 
+  updated(changedProperties) {
+    if (changedProperties.has('hasBatchActions')) {
+      this.setAttribute('tabindex', `${this.hasBatchActions ? -1 : 0}`);
```
asudoh, line 30:
```suggestion
      this.setAttribute('tabindex', `${this.hasBatchActions ? '-1' : ''}`);
```

asudoh (commented):
Thanks for clarifying why you made the change @IgnacioBecerra!

asudoh (approved):
LGTM 👍 - Thanks @IgnacioBecerra for the update!

PR #582, opened 2021-02-11
fix(combobox): now reads options in JAWS

IgnacioBecerra:
***PR created for QA feedback***

### Related Ticket(s)
https://app.zenhub.com/workspaces/carbon-for-ibmcom-5d449f3642eb1962336cbe52/issues/carbon-design-system/carbon-for-ibm-dotcom/4889

### Description

JAWS can now read the options in ComboBox.

### Changelog

**Changed**

- `assistiveStatusText` now also indicating the current input


src/components/dropdown/dropdown.ts
```diff
@@ -311,7 +317,9 @@ class BXDropdown extends ValidityMixin(HostListenerMixin(FocusMixin(LitElement))
 
     const nextItemText = nextItem.textContent;
     if (nextItemText) {
-      this._assistiveStatusText = nextItemText;
+      this._assistiveStatusText = `${nextItemText}${
+        this._inputNode && this._inputNode.value ? `, Current input is: ${this._inputNode.value}` : ''
+      }`;
```
asudoh, line 322: This string interpolation has to be translatable, similar to: https://github.com/carbon-design-system/carbon-web-components/blob/v1.12.0/src/components/pagination/pagination.ts#L115-L119

src/components/combo-box/combo-box.ts
```diff
@@ -234,21 +231,6 @@ class BXComboBox extends BXDropdown {
...
-    return `${prefix}-combo-box-beingtoggled`;
-  }
-
-  /**
-   * The name of the custom event fired after this combo box item is toggled upon a user gesture.
-   */
-  static get eventToggle() {
-    return `${prefix}-combo-box-toggled`;
-  }
-
```
asudoh, line 251: Any reason of removal...?

asudoh (commented):
Thank you for the update @IgnacioBecerra!

src/components/combo-box/combo-box.ts
IgnacioBecerra: Ah, I must've removed it by accident when pulling the changes from upstream. Will add it back!

src/components/dropdown/dropdown.ts
IgnacioBecerra: Could you expand on what it means for the string interpolation to be translatable? Thanks Akira!

asudoh: Thank you for asking; Some helpful resources are:
https://www.infragistics.com/community/blogs/b/devtoolsguy/posts/l10n-and-i18n-best-practices
https://kazupon.github.io/vue-i18n/guide/messages.html#named-formatting

src/components/combo-box/combo-box.ts
```diff
@@ -152,6 +152,7 @@ class BXComboBox extends BXDropdown {
         .value=${filterInputValue}
         role="combobox"
         aria-label="${inputLabel}"
+        aria-labelledby="assistiveStatus"
```
asudoh, line 155: Sounds that `aria-label` above is no longer used...?

asudoh (commented):
Thank you for the update @IgnacioBecerra!

src/components/combo-box/combo-box.ts
IgnacioBecerra: You're right, I'll remove it

asudoh: Another question. `<div id="assistiveStatus">` has `aria-live="assertive"`, which means screen reader should read it whenever the content changes. Did you mean it was not the case?

IgnacioBecerra: The `aria-live` content was being read out normally using Mac Voiceover, but without `aria-labelledby` the content wasn't being read out using JAWS for some reason.

asudoh: I see, would you try taking deeper look at how JAWS works with `aria-live`? You can try https://www.w3.org/TR/wai-aria-practices/examples/alert/alert.html for example. Some ARIA site suggests that `role="alert"` or `role="status"` in place of `aria-live`: https://www.w3.org/TR/WCAG20-TECHS/ARIA19.html#ARIA19-ex1 https://www.w3.org/WAI/WCAG21/working-examples/aria-role-status-searchresults/ - Would you try those, too? Thanks!

IgnacioBecerra: The current `<div id="assistiveStatus">` currently has `role="status"` along with `aria-live`. Removing `aria-live` still makes it work, but `aria-labelledby` is still needed for it to be read out.

asudoh: My apologies I was not clear enough; The links above has code examples. Would you try those code examples with JAWS to see how `aria-live`, `role="alert"` and `role="status"` work? Thanks!

IgnacioBecerra: From those examples, JAWS reads out the message for a split second then returns to what it was previously reading. It could be because I'm running JAWS in a Virtual Machine that it behaves that way, but I'm not sure. The Dropdown/Combobox changes worked alright on the VM, but I usually asked Ram to test out changes to confirm they worked in his machine.

asudoh: I see, thanks. The next steps would be:

* Work with India team to understand what they seem from W3 ARIA examples (above)
* If the result is what we expect (JAWS read out stuffs as soon as we put the content in), reduce our code or enhance W3 code to see what the fundamental differences are

The bottom line: `aria-live`, `role="alert"` or `role="status"` should work without being referred from our combo box input.

asudoh:
Had a discussion with @IgnacioBecerra offline, we are implementing the requirement in `@cabron/ibmdotcom-web-components` package for now, until the team figures out the root cause of the issue. Closing for now, we can definitely reopen this once RCA completes.

Issue #579, opened 2021-02-04
[number-input] How to handle updates through the arrows?


Tansito:
## Summary

Checking the behaviour for the `number-input` through the [storybook](https://web-components.carbondesignsystem.com/?path=/story/components-number-input--default) I couldn't find a way to handle the update for the value when you click in the arrows.

When you modify the value through the box the `onInput` event is fired but when the modification occurs through the arrows I couldn't find a way to handle the change.

asudoh:
@Tansito Thank you for reporting! Such kind of form elements should fire a custom event similar to [what `<bx-search>` does](https://github.com/carbon-design-system/carbon-web-components/blob/v1.11.0/src/components/search/search.ts#L57-L74), but I see that `<bx-number>` misses that part. Meanwhile you can listen to bubbling `click` event and inspect [`event.composedPath()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath) to see if it comes from the arrow button.

RobertaJHahn:
Since this is an implementation issue and a work around is provided, Closing this issue.

asudoh:
Thank you for trying to tidy this up @RobertaJHahn. Wondering if we can keep this open for a while and converting this as a feature request. I see that a fundamental feature is missing. Thanks!

RobertaJHahn:
Reopened and moved to the `To be triaged` pipeline to be reviewed and prioritized,

PR #589, opened 2021-02-16
feat(number-input): add custom event

asudoh:
### Related Ticket(s)

Refs #579.

### Description

Adds `bx-number-input` custom event to `<bx-number-input>` so that application code can listen to change in value, even if the change is done by our custom up/down button.

### Changelog

**New**

- `bx-number-input` custom event to `<bx-number-input>`.


kennylam (approved):
LGTM thanks @asudoh!

Issue #587, opened 2021-02-15
Data table: The position of selection checkbox is off


asudoh:
## Detailed description

> Describe in detail the issue you're having.

See the steps to reproduce.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

A bug.

> Is this issue related to a specific component?

Yes, data table.

> What browser are you working in?

Latest Chrome

> What version of Carbon Web Components are you using?

`1.11.1`

## Steps to reproduce the issue

1. Go to http://web-components.carbondesignsystem.com/?path=/story/components-data-table--sortable
1. Check off `has-selection` knob

BUG: You'll see the position of check box is off. Also, application cannot fix the issue because the check box is styled in Shadow DOM and there is no shadow part is defined for the check box.

PR #590, opened 2021-02-16
fix(data-table): fix position of check box

asudoh:
### Related Ticket(s)

Refs #587.

### Description

Fixes the wrong position of check box in `<bx-table-row>`. Also adds `part` attribute to the violating `<div>` as well as the `<input>` so application has control over their styles.

### Changelog

**New**

- `part` attribute to the container `<div>` of the table selection check box as well as the check box itself.

**Changed**

- Fixes the wrong position of check box in `<bx-table-row>`.


annawen1 (approved):
👍

PR #596, opened 2021-03-02
fix(link): update _handleClick() interface

asudoh:
### Related Ticket(s)

Refs carbon-design-system/carbon-for-ibm-dotcom#5322.

### Description

Updates `_handleClick()` method of `<bx-link>` so an inherited class can use the 1st arg, which is the event.

### Changelog

**Changed**

- `_handleClick()` method of `<bx-link>` so an inherited class can use the 1st arg, which is the event.


jeffchew (approved):
LGTM!

PR #599, opened 2021-03-04
chore(build): remove automerge.yml

asudoh:
### Description

`automerge.yml`, originally written for `carbon-for-ibm-dotcom` merges commits from `release/*` branch to `master`. OTOH,
`carbon-web-components` release process cherry-picks commits from `master` to `release/*`, which means, the merge direction is the way around.

Because of that nature, we don't need to merge `release/*` to `master`, and thus the CI script, `automerge.yml`, fails. This change removes `automerge.yml` therefore.

### Changelog

**Removed**

- `automerge.yml`


IgnacioBecerra (approved):
Thanks for the explanation! The removal makes sense to me.

PR #608, opened 2021-03-23
chore(dependencies): upgrade Carbon to `v10.31.x`

asudoh:
### Changelog

**Changed**

- Upgrades Carbon to `v10.31.x`.

Issue #586, opened 2021-02-15
UI Shell: Divider in side bar


asudoh:
### The problem

Carbon core is implementing a divider in aide nav.

### The solution

Implement it.

PR #609, opened 2021-03-23
feat(ui-shell): introduce divider in side bar

asudoh:
Depends on #608 - Until #608 is merged, the demo won't work.

### Related Ticket(s)

Refs #586.

### Changelog

**New**

- `<bx-side-bar-divider>`, a divider in side bar.


kennylam (approved):
Thanks for this @asudoh!

Issue #577, opened 2021-01-26
Implement short tag size


asudoh:
### The problem

Carbon core is implementing short tag size variant.

### The solution

Implement it.

### Additional information

As this point, `.bx--tag--sm` seems to define the new CSS ruleset.

PR #610, opened 2021-03-23
feat(tag): add size variant

asudoh:
Depends on #608 - Until #608 is merged, the demo won't work.

### Related Ticket(s)

Refs #577.

### Changelog

**New**

- `size` attribute in `<bx-tag>` and `<bx-filter-tag>`.


kennylam (approved):
LGTM thanks @asudoh!

Issue #498, opened 2020-09-26
Tabs: Add light variant


asudoh:
### The problem

Carbon core recently implemented `light` variant of tabs.

### The solution

Implement it.

RobertaJHahn:
@asudoh Just cleaning up the issues related to the Carbon Web Components January 13 release.  This issue is included in the release, but doesn't appear to be finished.  Should I move this issue to the next release in March?

cc:  @jeffchew

RobertaJHahn:
@asudoh I just found my notes on the scheduling of this issue.  Moving it to the March release.  

cc:  @jeffchew

PR #611, opened 2021-03-24
feat(tabs): add light variant

asudoh:
### Related Ticket(s)

Refs #498.

### Description

Adds `color-scheme="light"` option to `<bx-tabs>`.

Also:

* Removes top margin from regular tab panels demo as the design seems no longer dictates it.
* Adds tab panels to Angular/Vue stories.

### Changelog

**New**

- Support for `color-scheme="light"` in `<bx-tabs>`.
- Tab panels to Angular/Vue stories.

**Removed**

- The top margin from regular tab panels demo.


src/components/tabs/tabs-story-vue.ts
```diff
@@ -1,17 +1,28 @@
...
+const injectStoryStyle = () => {
+  const found = document.querySelector('style#tabs-story');
+  if (!found) {
+    const style = document.createElement('style');
+    style.id = 'tabs-story';
+    style.textContent = styles.cssText;
+    document.head.appendChild(style);
+  }
+};
+
```
kennylam, line 25: Looks good thanks @asudoh! Just a small question, why is manual injection necessary for the styles?

asudoh: @kennylam Good question; IIRC Vue template doesn't allow `<style>` tag, and thus above code takes raw DOM approach. In short, we need to create story-specific style.

kennylam: Got it, thanks @asudoh!

kennylam (approved):

PR #615, opened 2021-04-02
chore(dependencies): upgrade `sass-loader`

asudoh:
### Description

Upgrades `sass-loader` to the latest version that support WebPack4, so that we can use `webpackImporter: false` option  to disable `sass-loader`'s custom Sass importer that was one of the bottlenecks of Storybook build.

Also introduces `cache-loader` in there to further boost the performance.

Quick benchmark result is like:

* `webpackImporter: true`: 37s
* `webpackImporter: false`: 23s
* Using `cache-loader`: 16s

Also upgrades `css-loader`, `node-sass` and `style-loader`.

### Changelog

**Changed**

- Upgrades `css-loader`, `node-sass`, `sass-loader` and `style-loader`.
- `sass-loader` to use `webpackImporter: false` option as well as usage of `cache-loader` with it.


kennylam (approved):
Nice one thanks @asudoh!

PR #622, opened 2021-04-05
fix(modal): fixes modal tab index

andysherman2121:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/5060

### Description

The `bx-modal` had a typo for `tab-index`. This was causing an issue when JAWS looks for the modal dialogue. `leaving-ibm-modal` already had an `aria-labelledby` for the modal dialogue but it was looking at the wrong one.


asudoh (approved):
Good catch - Thanks @andysherman2121!

Issue #620, opened 2021-04-05
Combobox does not reset cleared input after losing focus


grommett:
## Detailed description

This issue is related to the `ComboBox` component.

When a combobox has a selection and the user clears the input (clicks 'x') and the component loses focus, it does not reset to its previous value.

## Steps to reproduce the issue

1. Load up a Combobox component with a few options and one default selected.
1. Clear the input by clicking the 'x'
1. Type nothing
1. Click off the combo box so that it loses focus

**Result:** The combobox remains shows its `trigger-content` value or empty if its `trigger-content` is not set.
**Expected:** The combobox should reset to its previously selected value.

## Additional information
- Pen for this with steps to repeat included. https://codepen.io/grommett/pen/zYNzvmq?editors=1111

PR #624, opened 2021-04-09
fix(combo-box): make X button clear the selection

asudoh:
### Related Ticket(s)

Refs #620.

### Description

The X button in `<bx-combo-box>` used to merely clear the type-ahead field, which controls which `<bx-combo-box-item>` should be highlighted.

However, X button in the React version clears the selection as well. This change makes such behavior in sync with the React version.

The type-ahead field of `<bx-combo-box>` controls the highlight of `<bx-combo-box-item>`, not selection, and this change keeps such behavior intact. In other words, this change affect only the behavior of the X button.

### Changelog

**Changed**

- X button in `<bx-combo-box>` now clears the selection in addition to the highlight.


jeffchew (approved):
LGTM!

PR #629, opened 2021-05-04
test(integration-build): temporarily suppressing test integration build

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticed that the integration build tests are failing, which was also
something noticed in the `Carbon for IBM.com` repo around the same time,
 and none of the recent code changes would indicate something that would
  cause this. This is possibly an environment related issue (possible
  latest Chrome version) that might be causing some rendering issues to
  occur within these tests, but would need to dig more into this in a
  separate PR.

### Changelog

**Changed**

- Removed `test:integration:build` from `ci-check` script


IgnacioBecerra (approved):

asudoh:
Weighing in; From [recent CI log](https://github.com/carbon-design-system/carbon-web-components/runs/2501028722), suggested that the team ensures React SSR and Next CodeSandbox examples are not broken.

PR #648, opened 2021-06-07
fix(dropdown): added formdata event support

IgnacioBecerra:
### Related Ticket(s)
None.

### Description
`<bx-dropdown>` currently lacks FormData support that is seen in other component such as `<bx-input>`, meaning that the selected Dropdown value wouldn't get passed upon Form submission. This PR adds support for said feature.

### Changelog

**New**

- added support for `FormData` event in `<bx-dropdown>`


jeffchew (approved):
LGTM, might want a second pair of eyes just in case though!

Issue #647, opened 2021-06-03
Carbon Slider out of range when min != 0


tambien:
## Detailed description

When the slider's min value is not 0, it goes out of the edges of the slider range. 

Here's a demo https://codesandbox.io/s/carbon-slider-issue-c7dsv

## Steps to reproduce the issue

1. Create a `<bx-slider>`
2. Set `min` to anything that's not 0
3. Adjust the slider range to the max value

## Additional information

![image](https://user-images.githubusercontent.com/1323402/120714795-b15e9580-c491-11eb-8209-cb5cbe311e1a.png)

Tested on latest Chrome

asudoh:
Wondering if the team wants to subtract `min` from `value` before dividing it by the range, at [`.rate` getter](https://github.com/carbon-design-system/carbon-web-components/blob/v1.13.3/src/components/slider/slider.ts#L86-L90). The repo welcomes contributions, too @tambien!

PR #649, opened 2021-06-08
fix(slider): fixes out of range positioning

kennylam:
### Related Ticket(s)

#647

### Description

Updates the slider button position correctly when a minimum value other than 0 is given.

### Changelog

**Changed**

- subtracts `min` from overall value (thanks @asudoh!)


IgnacioBecerra (approved):
Tested and works fine now, thanks @kennylam!

PR #650, opened 2021-06-08
fix(select): show proper attributes in readme

IgnacioBecerra:
### Related Ticket(s)
None

### Description
The properties of `<bx-select-item>`'s were rendered twice by accident -- fixed it so `<bx-select>` properties render in the docs.

### Changelog

**New**

- showing `<bx-select>` properties in README


NSDrowned (approved):
LGTM!

PR #659, opened 2021-06-14
feat(button): add isExpressive prop

ariellalgilmore:
### Related Ticket(s)

#3982

### Description

add isExpressive prop

### DIFF
--- a/src/components/button/button.ts
+++ b/src/components/button/button.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2020
+ * Copyright IBM Corp. 2019, 2021
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -85,6 +85,12 @@ class BXButton extends FocusMixin(LitElement) {
   iconLayout = BUTTON_ICON_LAYOUT.REGULAR;
 
   /**
+   * `true` if expressive theme enabled.
+   */
+  @property({ type: Boolean, reflect: true })
+  isExpressive = false;
+
+  /**
    * Button kind.
    */
   @property({ reflect: true })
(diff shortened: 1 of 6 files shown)


src/components/button/button.ts
```diff
@@ -156,8 +163,11 @@ class BXButton extends FocusMixin(LitElement) {
       [`${prefix}--btn--${kind}`]: kind,
       [`${prefix}--btn--disabled`]: disabled,
       [`${prefix}--btn--icon-only`]: hasIcon && !hasMainContent,
-      [`${prefix}--btn--${size}`]: size,
+      [`${prefix}--btn--sm`]: size === 'sm' && !isExpressive,
+      [`${prefix}--btn--xl`]: size === 'xl',
+      [`${prefix}--btn--field`]: size === 'field' && !isExpressive,
       [`${prefix}-ce--btn--has-icon`]: hasIcon,
+      [`${prefix}--btn--expressive`]: isExpressive,
```
annawen1, line 170: Do we use this class name to style at all? Seems like we only use the attribute to style

IgnacioBecerra (approved):
LGTM, thanks @ariellalgilmore!!

jeffchew:
LGTM! One thing though, noticed that the `isExpressive` knob isn't working in the react wrapper deploy preview. I didn't check the other expressive components though.

src/components/button/button.ts
ariellalgilmore: Screen Shot 2021-06-15 at 10 04 45 AM

yeah the class name is being styled from carbon core

ariellalgilmore:
> LGTM! One thing though, noticed that the `isExpressive` knob isn't working in the react wrapper deploy preview. I didn't check the other expressive components though.

oh didn't touch the other stories. Will add! thanks :)

src/components/button/button.ts
annawen1: ohh ok gotcha, cool!

annawen1 (approved):
LGTM!

PR #666, opened 2021-07-08
chore(eslint): carbon upgrade 10.38.0 eslint

kennylam:
### Related Ticket(s)

Duplicate of #663 with an additional fix to `.eslintignore`. This will prevent linting the `lib` directory which is causing `ci-check` to fail.

### Changelog

**New**

- upgrade Carbon to v10.38.0

**Changed**

- add `lib` to `.eslintignore`


jeffchew (approved):
Thanks @kennylam !

Issue #661, opened 2021-06-15
Provide carbon-web-components via CDN URL's


rkgriff:
### The problem

Need the ability to source carbon-web-components directly without creating module bundle files.  

### The solution

Make the carbon-web-components available via CDN (content delivery network) URL's.

### Additional information

Need to support a legacy application using carbon-web-components and cannot easily use module bundles.

rkgriff:
Thanks team.  Great news.

PR #681, opened 2021-08-10
feat(import): introduce simplified entry points

jeffchew:
### Related Ticket(s)

Refs #661 

### Description

This introduces a singular entry point for each component, which would
simplify importing components holistically. In addition, this is a
prerequisite to the generation of CDN artifacts for each components so
that adopters can use the CDN urls instead of including as part of their
 application build process.

### Changelog

**New**

- Singular component level imports

### DIFF
--- a/src/components/accordion/index.ts
+++ b/src/components/accordion/index.ts
@@ -1,0 +1,11 @@
+/**
+ * @license
+ *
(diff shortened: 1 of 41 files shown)


andysherman2121 (approved):
LGTM

PR #685, opened 2021-08-16
docs(cdn): adding documentation for CDN artifacts

jeffchew:
### Related Ticket(s)

Refs #661 

### Description

This adds documentation for the main README on how to use in lieu of importing via ES modules.

### Changelog

**Changed**

- `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,113 @@ The effort stems from https://github.com/carbon-design-system/issue-tracking/iss
 
 ## Getting started
 
+### Using CDN
+
+#### How to install
+
+All components are available via CDN. This means that they can be added to your application without the need of any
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -52,6 +57,117 @@ The effort stems from https://github.com/carbon-design-system/issue-tracking/iss
 
 ## Getting started
 
+### Using CDN
+
+#### How to install
+
+All components are available via CDN. This means that they can be added to your application without the need of any
```
emyarod, line 64: not sure which is preferable here

```suggestion
All components are available via CDN. This means that they can be added to your application without the need for any
```

emyarod (approved):
looks good to me!

PR #687, opened 2021-08-20
Use `import type` from TypeScript

abdonrd:
### Description

Use `import type` from TypeScript.

### Changelog

**Changed**

- Use `import type` from TypeScript.


jeffchew (approved):
Thank you @abdonrd !

Issue #558, opened 2020-12-10
[Modal] focus gets trapped on close


frodesigns:
When dealing with form inputs within a bx-modal and clicking our Submit button (`<bx-btn data-modal-close>` triggers the beforeclose event), validations are run, preventDefault is called for errors, and focus becomes trapped on the Submit button and the modal close X button, and we are unable to click into inputs to re-focus.  This has been observed with both the bx-input and bx-date-picker.

I've narrowed it down that I don't think any of our project's code, such as the validation, is causing this. It must be something in the bx-modal code related to the beforeclose or close events.

asudoh:
Hi @frodesigns thank you for reporting! Tried to create a content like below, but couldn't reproduce the problem. Would you try forking https://codesandbox.io/s/github/carbon-design-system/carbon-web-components/tree/master/examples/codesandbox/basic, put your content and see if the problem reproduces? If so, could you please share the fork? Some movie shots of the user interaction you described will also be a help. Thanks!

```html
<bx-modal id="modal">
  <bx-modal-header>
    <bx-modal-close-button></bx-modal-close-button>
    <bx-modal-label>Label (Optional)</bx-modal-label>
    <bx-modal-heading>Modal Title</bx-modal-heading>
  </bx-modal-header>
  <bx-modal-body>
    <form>
      <bx-form-item>
        <bx-date-picker>
          <bx-date-picker-input kind="single" label-text="Date picker">
          </bx-date-picker-input>
        </bx-date-picker>
      </bx-form-item>
      <bx-form-item>
        <bx-input id="input" label-text="input"></bx-input>
      </bx-form-item>
    </form>
  </bx-modal-body>
  <bx-modal-footer>
    <bx-btn kind="secondary" data-modal-close>Cancel</bx-btn>
    <bx-btn kind="primary">Save</bx-btn>
  </bx-modal-footer>
</bx-modal>
<script type="module" src="src/index.js"></script>
<script type="text/javascript">
  document.addEventListener('click', event => {
    if (event.target.id === 'launcher') {
      document.getElementById('modal').open = true;
    }
  });
  document.addEventListener('bx-modal-beingclosed', event => {
    const input = document.getElementById('input');
    if (!input.value) {
      event.preventDefault();
    }
  });
</script>
```

frodesigns:
I also can't reproduce this way, may have to look at another approach.

asudoh:
[...]

PR #689, opened 2021-08-23
fix(modal): include select shadow root children in focus wrap queries

emyarod:
### Related Ticket(s)

#558

### Description

This PR updates the `currentContains` queries to include all shadow root children (except for the `endSentinel` in order to preserve expected focus wrap behavior. unsure if the `startSentinel` needs to be an exception as well)

observe that clicking on the space in the modal body between the copy and the footer buttons no longer causes focus to be automatically placed on the modal close button any longer, and keyboard focus wrap behavior remains unchanged

#### original behavior (focus is returned to close button after clicking on the empty space):

https://user-images.githubusercontent.com/8265238/130491057-e1c09efe-1637-4876-8cf9-5153df93fbfe.mov

#### updated behavior (focus is not moved to the close button after clicking on the empty space):

https://user-images.githubusercontent.com/8265238/130491060-0e325458-6ad7-412f-a8b4-e74fc373d144.mov

### Changelog

**Changed**

- modal `currentContains` logic in `handleBlur` method


jeffchew (approved):
LGTM!

Issue #696, opened 2021-09-03
Style discrepancies with core Carbon buttons


emyarod:
## Detailed description

> Describe in detail the issue you're having.

There seem to be a few style discrepancies between the web component and core Carbon button components

the shared styles between icon-only buttons and buttons with text and icons also leads to some unintended spacing issues when the expressive variant is used

related downstream issue https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/6202

## Additional information

for example icon-only buttons seem slightly off center among other spacing discrepancies

![image](https://user-images.githubusercontent.com/8265238/132033349-acbdc07b-4572-4f07-81c6-dd7e33ef90f0.png)
![image](https://user-images.githubusercontent.com/8265238/132033652-1543ef57-1b0f-4fc1-ab64-3aee3c08a4e8.png)

PR #697, opened 2021-09-03
fix(button): follow core Carbon icon positioning

emyarod:
### Related Ticket(s)

#696

### Description

This PR updates the button styles to more closely match the styles from core Carbon and apply them to the web component version. Confirm that all button variants (default, icon, text + icon) appear as expected (including expressive versions)

image

original:

image

updated:

image

core Carbon (expressive):

image


oliviaflory:
@emyarod I noticed the ghost and ghost danger buttons have a slight issue where the icon and text overlap in the text + icon variants:
Screen Shot 2021-09-03 at 10 00 43 AM
Screen Shot 2021-09-03 at 10 03 41 AM


Otherwise looks great!

emyarod:
@oliviaflory the ghost variants should be fixed now, also added a fix for the condensed icon layout!

oliviaflory:
@emyarod thanks for correcting the overlap!

My one question is about the condensed versions of the ghost buttons

In the primary button, the regular is spaced out with ~64px between the text and icon, and the condensed seems to be ~ 4 or 8px space between text and icon

**Regular**
Screen Shot 2021-09-08 at 11 44 30 AM
**Condensed**
Screen Shot 2021-09-08 at 11 43 08 AM


In the ghost button version, they seem to be the same amount of space

**Regular**
Screen Shot 2021-09-08 at 11 44 18 AM

**Condensed**
Screen Shot 2021-09-08 at 11 43 59 AM

I can't see any knobs to change from regular to condensed in the [Carbon react storybook](https://react.carbondesignsystem.com/?path=/story/components-button--ghost), so I'm not sure what the expected spacing is, would you mind double checking? If this is the same as the Carbon one, lmk and I can approve!

emyarod:
@oliviaflory the condensed sizing seems to be exclusive to `carbon-web-components`, it was introduced here https://github.com/carbon-design-system/carbon-web-components/pull/468 for better support for the data table's action bar in core Carbon. If this is something we don't want to display in our storybook I can remove that option

shixiedesign:
Hey @emyarod @oliviaflory , looking into this a bit, it seems like Carbon react handles ghost button styles differently than Carbon web components. Is it simply that WC's ghost button styling need an update? Ideally we don't introduce new variables  .

Sep-08-2021 13-58-47

Sep-08-2021 13-59-38

emyarod:
@shixiedesign @oliviaflory sounds like it would be better to remove the `condensed` option from the storybook and keep it internal to WC then. we can also separate the work for aligning the ghost button implementations from this PR so that the downstream expressive modal work can continue in the meantime

emyarod:
@oliviaflory @shixiedesign the ghost buttons should be resolved now. I left the `icon-layout` knob in Storybook, but it's specific to `carbon-web-components`. it just has no effect on the ghost buttons (but will have effects on the other button types)

jeffchew (approved):
LGTM!

oliviaflory (approved):
thanks @emyarod 👍

Issue #695, opened 2021-08-30
[Radio button] Disabled radios are still clickable and emit changed event


frodesigns:
## Steps to reproduce the issue

1. Visit https://web-components.carbondesignsystem.com/?path=/story/components-radio-button--default
2. Go to Knobs and check Disabled
3. Go to Actions
4. Click a radio button
5. Radio UI selection changes and changed event is triggered

kennylam:
These are Storybook actions only. `onClick` needs to be set when authoring a page, whether the component is disabled or not.

frodesigns:
Storybook actions are representations of actual events triggering. This is a problem in production applications. Buttons should not be clickable when disabled, nor should they fire the changed event.

RobertaJHahn:
At a minimum, the buttons should not be clickable when disabled.  More discussion is needed around whether an event is logged when clicked.  ![image.png](https://images.zenhubusercontent.com/5d8b794c2134370001ce8d1a/a5008123-b729-46f9-bf4c-e6eda9a91619)

PR #703, opened 2021-09-29
fix(radio-button): apply `disabled` prop to input

emyarod:
### Related Ticket(s)

#695

### Description

This PR passes the value of the `disabled` prop to the shadow DOM `input` and also prevents `checked` from being applied if the input is already disabled

Confirm that a disabled radio input can no longer be manipulated


jeffchew (approved):
LGTM!

PR #712, opened 2021-10-11
fix(button): prevent event propagation to children of disabled buttons

### LINKED ISSUE #693: [Button] Disabled buttons emit click event - DDS consulting
## Description

Click event is triggered on disabled buttons when clicking the button text.

## Steps to reproduce the issue

1. Visit https://web-components.carbondesignsystem.com/?path=/story/components-button--default
2. Check "disabled" in Knobs 
3. Open Actions tab
4. Click button background (no click event is triggered)
5. Click button foreground text (click event is triggered)

andy-blum:
### Related Ticket(s)

Fixes #693 

### Description

Adds a `hostlistener` to the button component that uses the `capture` phase to prevent click propagation to child elements

### Changelog

**Changed**

- `disabled` buttons prevent click propagation to their child elements


andy-blum:
cc: @kennylam

Issue #683, opened 2021-08-13
[Multi-select]: add filterable variant


annawen1:
Carbon core has multi-select filterable variant. This request is to sync carbon-web-components to include this variant

https://react.carbondesignsystem.com/?path=/story/components-multiselect--filterable

<img width="757" alt="Screen Shot 2021-08-13 at 11 34 02 AM" src="https://user-images.githubusercontent.com/54281166/129383349-51bb570a-d5b6-4c84-828e-e04eb1bcacef.png">

PR #716, opened 2021-10-23
feat(multi-select): introduce filterable mode

IgnacioBecerra:
### Related Ticket(s)
#683

### Description
Introduced a filterable variant of the Multi Select component that works as a ComboBox that filters and is able to select multiple options.

### Changelog

**New**

- new `filterable` variant for `bx-multi-select`


emyarod (approved):
looks good to me!

abdonrd:
@IgnacioBecerra @jeffchew seems like we need to add a new snapshot, right?

PR #729, opened 2021-11-02
chore: update checkout and setup-node actions

abdonrd:
Related with this update, the `setup-node` has a built-in option to cache:
https://github.com/actions/setup-node#caching-packages-dependencies

If you want to use it, I can add it in another PR.

### Changelog

**Changed**

- Update checkout and setup-node actions


jeffchew (approved):
thank you @abdonrd , looks good to me!

Issue #730, opened 2021-11-02
Installation does not work


abdonrd:
With `yarn install` I get this error:

<img width="1999" alt="Screenshot 2021-11-02 at 12 12 12" src="https://user-images.githubusercontent.com/1007051/139836643-52a7406a-67a6-48f1-bdd2-eb69f986e188.png">

```
warning " > tsickle@0.37.1" has incorrect peer dependency "typescript@~3.6.4".
[4/4] 🔨  Building fresh packages...
[12/47] ⡀ husky
[2/47] ⡀ fsevents
[15/47] ⡀ puppeteer
[11/47] ⡀ node-sass
error /Users/abdonrd/Work/Forks/carbon-design-system/carbon-web-components/node_modules/node-sass: Command failed.
Exit code: 1
Command: node scripts/build.js
Arguments: 
Directory: /Users/abdonrd/Work/Forks/carbon-design-system/carbon-web-components/node_modules/node-sass
Output:
Building: /Users/abdonrd/.nvm/versions/node/v14.17.5/bin/node /Users/abdonrd/Work/Forks/carbon-design-system/carbon-web-components/node_modules/node-gyp/bin/node-gyp.js rebuild --verbose --libsass_ext= --libsass_cflags= --libsass_ldflags= --libsass_library=
gyp info it worked if it ends with ok
gyp verb cli [
gyp verb cli   '/Users/abdonrd/.nvm/versions/node/v14.17.5/bin/node',
gyp verb cli   '/Users/abdonrd/Work/Forks/carbon-design-system/carbon-web-components/node_modules/node-gyp/bin/node-gyp.js',
gyp verb cli   'rebuild',
gyp verb cli   '--verbose',
gyp verb cli   '--libsass_ext=',
gyp verb cli   '--libsass_cflags=',
gyp verb cli   '--libsass_ldflags=',
gyp verb cli   '--libsass_library='
gyp verb cli ]
gyp info using node-gyp@3.8.0
gyp info using node@14.17.5 | darwin | x64
gyp verb command rebuild []
gyp verb command clean []
gyp verb clean removing "build" directory
gyp verb command configure []
gyp verb check python checking for Python executable "/usr/local/opt/python@3.9/bin/python3" in the PATH
gyp verb `which` succeeded /usr/local/opt/python@3.9/bin/python3 /usr/local/opt/python@3.9/bin/python3
gyp ERR! configure error 
[...]

PR #731, opened 2021-11-02
chore: regenerate lock file and offline cache

abdonrd:
### Related Ticket(s)

Fix https://github.com/carbon-design-system/carbon-web-components/issues/730

### Changelog

**Changed**

- Regenerate lock file and offline cache


abdonrd:
There are just a few issues with some types.

PR #732, opened 2021-11-02
chore: run yarn upgrade

abdonrd:
I just run: `yarn upgrade`


abdonrd:
Type error:

>Type 'TemplateResult' is not assignable to type 'string | TemplateResult'

PR #734, opened 2021-11-02
chore: setup actions/setup-node@v2 cache

abdonrd:
As mention #729, just enabled the actions/setup-node@v2 built-in cache.

Let's see how improves the CI time.


abdonrd:
Wow! Seems like we change from ±9min to ±4min!

PR #736, opened 2021-11-02
chore: setup prettier-plugin-package

abdonrd:


abdonrd:
Oops! Prettier is outdated!

PR #737, opened 2021-11-03
Update dev dependencies

abdonrd:
Update a lot of the dev dependencies.
Multiple commits to check with the CI.


abdonrd:
How can we check what happened to Angular deploy preview?
Everything else has worked!

jeffchew:
Some of the snapshots seem a little off now, I approved some of them which just had some minor spacing things but the accordion is missing a lower border now:

image

Also the breadcrumb is missing the dividers:

image

jeffchew:
> How can we check what happened to Angular deploy preview? Everything else has worked!

The angular one has been finnicky lately, for some reason needs more memory to run on our CI. Going to see if restarting the build will work. But I did notice some visual diffs.

abdonrd:
Wow! I don't quite understand how there can be visual differences if I have only updated dev dependencies.

jeffchew:
> Wow! I don't quite understand how there can be visual differences if I have only updated dev dependencies.

Strange indeed. I think the packages that might remotely be responsible are these but not sure...

image

abdonrd:
> Strange indeed. I think the packages that might remotely be responsible are these but not sure...
> 
> image

I just reverted this commit, let's see it it works...

jeffchew:
No diffs after the update, looks like that was the culprit?

image

abdonrd:
> No diffs after the update, looks like that was the culprit?
> 
> image

No, I removed the sass updates!

jeffchew (approved):
LGTM!

abdonrd:
@jeffchew updated again with master!

PR #726, opened 2021-11-01
chore(lit): moved lit-element/lit-html to core dependencies

jeffchew:
### Related Ticket(s)

Refs https://github.com/carbon-design-system/carbon-web-components/issues/723

### Description

This moved the `lit-element`/`lit-html` libraries to core `dependencies`. `peerDependencies` was also cleaned up to no longer include `carbon-components`, as that has already been managed in `dependencies` for awhile now.

The scripts were also alphabetized in `package.json`, and added some additional entries to `.gitignore`.

### Changelog

**Changed**

- `lit-element`/`lit-html` moved to `dependencies`
- trimmed down `peerDependencies`
- Added `.npmrc` and `yarn-error.log` to `.gitignore`
- alphabetized scripts in `package.json`
- `README.md` updated to reflect installation changes

PR #742, opened 2021-11-03
docs(welcome): update getting started install section

abdonrd:
Update getting started install section

Because this: https://github.com/carbon-design-system/carbon-web-components/pull/726


jeffchew (approved):
Good catch!

PR #744, opened 2021-11-04
chore: prevents actions from forks

abdonrd:
### Description

Prevents action from creating a PR on forks as they have no access to the tokens.


jeffchew (approved):
LGTM!

PR #751, opened 2021-11-05
chore(dependencies): update typescript

abdonrd:
Update TypeScript version


src/components/number-input/number-input.ts
```diff
@@ -94,7 +95,9 @@ export default class BXNumberInput extends BXInput {
   @query('input')
   protected _input!: HTMLInputElement;
 
-  _testValidity() {
+  // TODO: Fix this type error
+  // @ts-ignore
+  _testValidity(): NUMBER_INPUT_VALIDATION_STATUS | VALIDATION_STATUS {
```
abdonrd, line 100: We need to fix this...

PR #753, opened 2021-11-05
chore(github-actions): improve GA logic

Tansito:
### Description

Whit this change the check for forks in GAs is more dynamic.

As a bonus in Qiskit we are adding a check to avoid the execution for some GAs with drafts too `github.event.pull_request.draft == false` just in case you could be interested.


jeffchew (approved):
Nice change, thank you!

PR #758, opened 2021-11-07
chore(dependencies): update ^0. deps versions

abdonrd:
Update `^0.` deps versions to `~0.` since these versions are not usually stable.


abdonrd:
@jeffchew can your re-ren the deploy preview? Everything else seems to work.

PR #774, opened 2021-11-09
chore(dependencies): update eslint-config-airbnb-base

abdonrd:
Update `eslint-config-airbnb-base`.

- Removing `use strict` because `eslint-config-airbnb-base` update.

---

~~Update ESLint & Prettier.~~

~~I had to upgrade from `eslint-config-carbon-base` to` eslint-config-carbon` in order to upgrade Prettier to v2 and the ESLint dependencies.~~


jeffchew (approved):
LGTM!

PR #781, opened 2021-11-10
fix(github-actions): turn back on percy update base

jeffchew:
### Related Ticket(s)

No related issue

### Description

Noticed that the percy base wasn't getting updated, and saw that the rule added was causing this. This should re-enable percy base updating.

### Changelog

**Changed**

- `percy-update-base.yml` remove repo target check rule


andysherman2121 (approved):
LGTM

emyarod (approved):
looks good to me!

PR #806, opened 2021-11-16
chore(deps): update flatpickr

abdonrd:
Update flatpickr


package.json
```diff
@@ -270,7 +270,7 @@
     "zone.js": "~0.8.29"
   },
   "peerDependencies": {
-    "flatpickr": "4.6.1"
+    "flatpickr": "4.6.9"
```
abdonrd, line 273: @jeffchew is `flatpickr` expected to be a `peerDependencies`?

abdonrd: I also see that `carbon-components` has `flatpickr` as dependency.

jeffchew (approved):

Issue #815, opened 2021-11-18
Slider not work with negative min


korgan00:
### Description

**What did I expect to happen?**
Setting a negative min and a positive max, the thumb should be able to be dragged all over the slide area.

**What happened instead**
The thumb can be dragged over the negative values but stops in 0.

### Component(s) impacted

Slider

https://user-images.githubusercontent.com/4138279/142446434-9514ab21-7b1d-4d7a-8c0f-12e8033dd28e.mov

### Browser

Chrome

### Carbon Web Components version

1.18

### Severity

Severity 1 = The design is broken in a critical way that blocks users from completing tasks or damages the brand. Affects major functionality, no workaround.

### Application/website

learn.qiskit.org

### CodeSandbox example

https://web-components.carbondesignsystem.com/react/index.html?path=/story/components-slider--default

### Steps to reproduce the issue (if applicable)

Set the negative value to -100

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

PR #816, opened 2021-11-18
fix(slider): negative values are now supported

korgan00:
### Related Ticket(s)

fix: https://github.com/carbon-design-system/carbon-web-components/issues/815

### Description

Now you can set negative min values and move the thumb all over the slider line.

### Changelog

**Changed**

- fix slider negative min value.


abdonrd (approved):
Thanks @korgan00! 👏🏻

jeffchew (approved):
LGTM!

PR #831, opened 2021-11-19
chore(deps): downgrade @storybook/angular

abdonrd:
Downgrade `@storybook/angular` to try to fix the stories.


abdonrd:
@kennylam seems like it works again, right?

jeffchew (approved):
Seems good now, thank you @abdonrd !

PR #894, opened 2021-11-24
chore: remove custom skipLibCheck

abdonrd:
Remove custom `skipLibCheck`.

We should remove the `// @ts-ignore`, but meanwhile, at least we remove the `skipLibCheck`.


jeffchew (approved):
Thank you @abdonrd !

Issue #776, opened 2021-11-09
[Combo Box] Page scroll when typing in combo box


frodesigns:
### Description

When typing in a combo box, the page scrolls down and is jarring to the user.

 

### Component(s) impacted

This is reproducible in Storybook: https://web-components.carbondesignsystem.com/?path=/story/components-combo-box--default

Here is a video of the issue:
https://user-images.githubusercontent.com/6588808/141011615-699d0107-2b95-47d8-9c1c-80a79796956e.mp4

### Browser

Chrome

### Carbon Web Components version

Latest

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

Kyndryl Catalyst

### CodeSandbox example

Storybook

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

emyarod:
related https://github.com/carbon-design-system/carbon-web-components/issues/722

frodesigns:
Hello. I don't think this issue is actually related to storybook in any way.  Catalyst leadership is asking for an escalation to fix this since it is impacting the main UX of the application.

Issue #692, opened 2021-08-30
Bx-combo-box wrong behavior on filtering.


SoloveyDm:
## Detailed description

https://codesandbox.io/s/github/carbon-design-system/carbon-web-components/tree/master/examples/codesandbox/cdn

1. Scroll the page so that the bx-combo-box is at the bottom of the page (see screenshot 1).
2. Enter any character in the filtration area.
3. This will cause the whole page to scroll up (see screenshot 2).

![screenshot_1](https://user-images.githubusercontent.com/89772242/131315824-75c44a44-ab7e-4e24-97ed-c4c6d2cbbb7a.jpg)
![screenshot_2](https://user-images.githubusercontent.com/89772242/131315827-42fd91c7-d366-465b-a119-095a828fd0f3.jpg)

RobertaJHahn:
dup of #776.  Closing

PR #901, opened 2021-11-29
fix(combo-box): ensure only div scrolls upon input

IgnacioBecerra:
### Related Ticket(s)
#692, #776

### Description
This PR removes the `scrollIntoView()` function when an item is highlighted upon input, which caused an issue where the whole page would jump instead of just within the div. This uses a new method using `scrollTop` to ensure only the menu div scrolls to the selected item in question properly, only if it's out of view.

### Changelog

**New**

- new scroll method to ensure selected item highlight is seen within the div

**Removed**

- `scrollIntoView()` function from `combo-box`


jeffchew:
@IgnacioBecerra looks like some prettier issues, can you take a look? 

image

jeffchew:
@IgnacioBecerra looks like some tests are failing now:

image

jeffchew (approved):
Thank you @IgnacioBecerra !

Issue #754, opened 2021-11-05
[Codesandbox]: Create codesandbox examples for each individual component


jeffchew:
### The problem

There are general codesandbox examples currently, but one of the nice things about having individual examples for isolated components is that it would assist with end users that would like to know how to integrate the component(s) they are interested in in a sandbox environment. We currently do this on `Carbon for IBM.com`, and provide the CSB link for each component.

In addition, the integration and e2e tests are based on each individual CSB example in `Carbon for IBM.com`, which has proven to be very effective with upstream integration testing.

### The solution

Create a codesandbox example for each individual component in `carbom-web-components`, and provide the link in each individual component README.

- follow-on work is managed under issue #755 

### Application/website

Carbon Web Components

### Business priority

Medium Priority = upcoming release but is not pressing

### Examples

An example pull request in `Carbon for IBM.com`:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7594

We run a deploy preview of the codesandbox examples:
https://webcomponents-codesandbox.s3-web.us-east.cloud-object-storage.appdomain.cloud/deploy-previews/7594/index.html

Plus also run the e2e/integration tests:
https://github.com/carbon-design-system/carbon-for-ibm-dotcom/runs/4099677969?check_suite_focus=true

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

PR #908, opened 2021-12-13
feat(codesandbox): add examples

ariellalgilmore:
### Related Ticket(s)

#754 

### Description

Add codesandbox examples for each component.

Updated small issues with cdn example.

### Changelog

**New**

- component folder containing a codesandbox folder for each component

**Changed**

- text for `tag` from "This is not a tag" to "This is a tag"

**Removed**

- {{removed thing}}


emyarod (approved):
just have a couple of questions:

should we update `basic/.gitignore` to also ignore the `components` directory? that way they won't show up in the existing CSB example if we still want to keep that example https://codesandbox.io/s/github/carbon-design-system/carbon-web-components/tree/feat/codesandbox/examples/codesandbox/basic

also should we set a width of 300px on the container around each component like the current basic CSB example? https://github.com/carbon-design-system/carbon-web-components/blob/feat/codesandbox/examples/codesandbox/basic/index.html#L19-L21

other than that looks good to me!

ariellalgilmore:
> just have a couple of questions:
> 
> should we update `basic/.gitignore` to also ignore the `components` directory? that way they won't show up in the existing CSB example if we still want to keep that example https://codesandbox.io/s/github/carbon-design-system/carbon-web-components/tree/feat/codesandbox/examples/codesandbox/basic
> 
yeah good catch! will check with team if we still want to keep that example and if so then ignore the `components` directory
> also should we set a width of 300px on the container around each component like the current basic CSB example? https://github.com/carbon-design-system/carbon-web-components/blob/feat/codesandbox/examples/codesandbox/basic/index.html#L19-L21
> 
I wasn't sure about that either. I was looking at Carbon's react codesandbox and they didn't set the width. I'll check with our team as well on that!
> other than that looks good to me!

PR #920, opened 2022-01-06
docs(cdn): remove reference to next tag

jeffchew:
### Related Ticket(s)

No related issue

### Description

We currently do not have a `next` tag for CDN so removing references to it.

### Changelog

**Changed**

- Updated `storybook-cdn.ts` file to not include reference to `next` tag


emyarod (approved):
looks good to me!

Issue #904, opened 2021-12-10
File uploader: Cannot select same file again


tpluscode:
### Description

If I have a file uploader where `multiple=false`, it is not possible to reupload a file, say, if it failed an the user removes it

### Component(s) impacted

`<bx-file-drop-container>`

I see this happens because the event handled is `input@change` and obviously selecting the same file again is not a "change" per se.

### Browser

Chrome

### Carbon Web Components version

1.18

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

private

### CodeSandbox example

https://codesandbox.io/s/stupefied-goodall-sjngl?file=/index.html

### Steps to reproduce the issue (if applicable)

1. Select a file
2. Click `x` to remove
3. Select same file again

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/master/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

PR #924, opened 2022-01-12
fix(drop-container): clear input value on change

emyarod:
### Related Ticket(s)

#904

### Description

This PR clears the file uploader input `value` so that the same file can be reuploaded after being cleared from the previous upload attempt

To test this, upload a file, remove it, and then reupload the same file and confirm that it is added to the file list correctly


IgnacioBecerra (approved):
Confirmed that files can be reuploaded without error. LGTM, thanks @emyarod!

jeffchew (approved):
LGTM!

Issue #551, opened 2020-12-03
[Tooltip] Alignment option in interactive tooltip


asudoh:
### The problem

Carbon core is adding alignment option in interactive tooltip.

https://github.com/carbon-design-system/carbon/pull/7309

### The solution

Implement it.

oliviaflory:
Assuming this is the alignment of where the caret is on the tooltip, would be great to get this in, I think I've seen our team suggesting the alignment on our own components that would utilize tooltip.

https://react.carbondesignsystem.com/?path=/story/components-tooltip--default-bottom

<img width="1443" alt="Screen Shot 2022-01-10 at 11 37 02 AM" src="https://user-images.githubusercontent.com/45692622/148803123-c278ed33-ff01-4678-bd9b-c5b6c2f0b3ac.png">

emyarod:
we may also need to update with some follow up features:

https://github.com/carbon-design-system/carbon/pull/9556
https://github.com/carbon-design-system/carbon/pull/9606

PR #930, opened 2022-01-21
fix(tooltip): update to match component spec

emyarod:
### Related Ticket(s)

unblocks #551 so that the PR to address #551 can be exclusively scoped to the alignment property 

### Description

This PR adds support for activating and dismissing the tooltip via keyboard, and updates the positioning of the tooltip caret in the left and right directions. 

### Changelog

**New**

- keyboard support for activating/dismissing tooltip

**Changed**

- update tooltip copy type tokens
- tooltip caret positioning


ariellalgilmore:
Looks great!! one minor thing. I noticed for positioning left, top, and right the tooltip gets cut off. Can we center it more?
Screen Shot 2022-01-27 at 10 17 01 AM

emyarod:
@ariellalgilmore should be resolved now once the deploy previews are updated!

ariellalgilmore:
some tests are still failing :/ https://github.com/carbon-design-system/carbon-web-components/runs/5010076991?check_suite_focus=true

emyarod:
for some reason the test is timing out because of the CSS to center the component in storybook...

ariellalgilmore (approved):
LGTM!

oliviaflory:
@emyarod sorry we missed this, we will be sure to keep an eye on the Carbon web components repo moving forward!

1. The "Learn more" link in the default tool tip doesn't look like the correct color token, we could update in the next PR too. I think it should be `$inverse-link`
Screen Shot 2022-02-02 at 9 58 51 AM

2.  Are we expecting a knob for the caret alignment (start, center, end) in the Default tooltip story, like the ones in the definition tooltip?
Screen Shot 2022-02-02 at 9 58 02 AM
Screen Shot 2022-02-02 at 9 58 07 AM

Everything looks great!

emyarod:
gotcha @oliviaflory, yes the alignment knob is coming in a follow up PR and I can include fixes for the first bullet point in that as well!

PR #940, opened 2022-02-03
feat(tooltip): implement alignment prop

emyarod:
### Related Ticket(s)

#551
https://github.com/carbon-design-system/carbon-web-components/pull/930

### Description

This PR adds the `alignment` prop to the interactive tooltip. The prop can be tested in the component story via the `alignment` knob

image

### Changelog

**New**

- `alignment` prop in tooltip


jeffchew:
Thanks @emyarod !

Something seems to be up with the integration tests, can you take a look?

image

emyarod:
@jeffchew yeah for some reason when I change the storybook styles on the story container div it will fail (https://github.com/carbon-design-system/carbon-web-components/pull/940/files#diff-a02b3b8aee424d6c6675a6fc678b937d5dc81bd89ec68ef91c5838a9fe55eccb). I just centered the story for easier reviewing but I will remove that style if everything else looks good

jeffchew:
> @jeffchew yeah for some reason when I change the storybook styles on the story container div it will fail (https://github.com/carbon-design-system/carbon-web-components/pull/940/files#diff-a02b3b8aee424d6c6675a6fc678b937d5dc81bd89ec68ef91c5838a9fe55eccb). I just centered the story for easier reviewing but I will remove that style if everything else looks good
> 
> ran into this issue in an earlier PR [#930 (comment)](https://github.com/carbon-design-system/carbon-web-components/pull/930#issuecomment-1026052097)

So strange! I guess for now we can revert it, but wondering if it's easy to update the test so it can find it while centered instead...

emyarod:
yeah I'll have to take another crack at it since I wasn't able to figure out the reason while the previous PR was in review

emyarod:
I think we can separate the storybook wrapper div styling issue out to a different issue. I will undo the storybook style changes if the interactive tooltip alignment and direction props look good

oliviaflory:
This looks great @emyarod 

A small thing I noticed, the focus on the link looks a bit off in color, looks like it might be `$focus` and should be `$inverse-focus-ui` on top of the high contrast background here.

**Current**
Screen Shot 2022-02-16 at 4 34 55 PM
**Expecting**
Screen Shot 2022-02-16 at 4 38 43 PM

emyarod:
@oliviaflory should be updated now! on a side note it looks like tooltips might not contain interactive content within them any more based on https://carbon-components-react.netlify.app/?path=/story/components-tooltip--default-bottom

it might be due to the new popover/toggletip and popover components due to land in core Carbon

oliviaflory:
> @oliviaflory should be updated now! on a side note it looks like tooltips might not contain interactive content within them any more based on https://carbon-components-react.netlify.app/?path=/story/components-tooltip--default-bottom
>
> it might be due to the new popover and toggletip components due to land in core Carbon

@emyarod if it's not available anymore, do you think we should keep it in Carbon Web Components or is this an example where we'd remove it to match Carbon React?

I saw the link and button options in "uncontrolled tooltip" but not sure what that means exactly.
https://carbon-components-react.netlify.app/?path=/story/components-tooltip--uncontrolled-tooltip

emyarod:
@oliviaflory yep I'm in favor of removing the example content to match! as for the uncontrolled tooltip example, that story is for demonstrating some React-specific approaches, so it's not relevant to the web components implementation

oliviaflory (approved):
@emyarod thanks for the update, looks good!

ariellalgilmore (approved):

PR #942, opened 2022-02-04
docs(input): fix input type knob

emyarod:
### Related Ticket(s)

#87

### Description

This PR updates the input `type` knob generator so that it works properly. Changing the input `type` prop should now change the rendered input type as expected


jeffchew (approved):
LGTM!

Issue #946, opened 2022-02-08
Sync component sizes with core Carbon


emyarod:
### Description

https://github.com/carbon-design-system/carbon/issues/8151
https://github.com/carbon-design-system/carbon/issues/8370

### Component(s) impacted

all form components

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon Web Components version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

.

### CodeSandbox example

https://web-components.carbondesignsystem.com/?path=/story/introduction-welcome--page

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

kennylam:
Related to component sync effort in https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8303

kennylam:
@emyarod Need to confirm how v11 will dictate this work, as some of the changes may not be 'live' until v11 is released.

Issue #945, opened 2022-02-08
Icon-only button missing tooltips on hover/focus


emyarod:
### Description

the icon-only button does not display a tooltip on hover or focus

<img width="167" alt="image" src="https://user-images.githubusercontent.com/8265238/153068942-22431ea7-c3c9-4a7b-a547-073f2c68bedb.png">

### Component(s) impacted

button
code snippet
password input
possibly others

### Browser

Chrome, Safari, Firefox, Microsoft Edge

### Carbon Web Components version

canary

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

-

### CodeSandbox example

https://web-components.carbondesignsystem.com/?path=/story/components-button--icon

### Steps to reproduce the issue (if applicable)

1. visit https://web-components.carbondesignsystem.com/?path=/story/components-button--icon
2. hover over or focus on the component
3. observe that no tooltip appears

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

Issue #305, opened 2020-02-04
Component: Password input


asudoh:

oliviaflory:
Useful for teams with any type of login or password (landing pages, URX)
Would be nice to have feature parody

Seem to have some naming differences between [Carbon React](https://react.carbondesignsystem.com/?path=/story/components-textinput--default) and [Web components](https://web-components.carbondesignsystem.com/?path=/story/components-input--default) storybooks
- `TextInput` vs `Input`

emyarod:
partially blocked by #945

PR #947, opened 2022-02-08
feat(input): implement password input

emyarod:
### Related Ticket(s)

#305
#942
#946

### Description

This PR updates the input component so that input `type="password"` together with the `show-password-visibility-toggle` being set to `true` will enable the Carbon password input styles

Confirm that the password input appears and functions as expected (Note that the tooltip on hover and focus of the password visibility toggle button is blocked by #945)

image

image

image

image

### Changelog

**New**

- `show-password-visibility-toggle` prop

**Changed**

- default input field size (`md` instead of `lg` now)


RichKummer (approved):
Looks good @emyarod!

kennylam (approved):
Thanks @emyarod!

PR #962, opened 2022-02-17
chore(deps): downgrade `execa` to v5

kennylam:
### Description

This downgrades `execa` to v5, as v6 was causing integration build tests to fail.

### Changelog

**Changed**

- drop `execa` to v5


emyarod (approved):
looks good to me! just the one tooltip integration test failure which seems to be a preexisting issue. thanks!

emyarod (commented):
it looks like something in our integration test scaffolding is generating untracked changes

image

emyarod (approved):
going to reapprove since the untracked files are an existing issue which can be addressed separately

PR #964, opened 2022-02-21
fix(jest): downgrade to v26

ariellalgilmore:
### Related Ticket(s)

none

### Description

should resolve unit test failures 

### Changelog

**Changed**

- downgraded jest to 26


emyarod (approved):
(reverting https://github.com/carbon-design-system/carbon-web-components/pull/886)

looks good to me!

PR #983, opened 2022-03-10
fix(deps): downgrade mkdirp fixes storybook:react

kennylam:
### Description

This downgrades the `mkdirp` package. The previous upgrade of this package introduced breaking changes to the React Storybook environment, preventing it from loading.

### Changelog

**Changed**

- downgrade `mkdirp` to previous version


emyarod:
reverts https://github.com/carbon-design-system/carbon-web-components/pull/864

emyarod (approved):
looks good to me! would we be able to go back to v0.5.5 or does it need to be on v0.5.0?

kennylam:
> looks good to me! would we be able to go back to v0.5.5 or does it need to be on v0.5.0?

Oh good call, I didn't even notice the previous one had a patch version. Once it worked I was like commit! But I'll bump it back to the correct previous version.

Issue #970, opened 2022-03-01
Keyboard navigation for Tabs not working


emyarod:
### Description

it is not possible to navigate the tabs component with the keyboard

### Component(s) impacted

tabs

### Carbon Web Components version

canary

### Severity

Severity 2 = Aspects of design is broken, and impedes users in a significant way, but there is a way to complete their tasks. Affects major functionality, has a workaround.

### Application/website

n/a

### CodeSandbox example

https://web-components.carbondesignsystem.com/?path=/story/components-tabs--default

### Steps to reproduce the issue (if applicable)

try to navigate the tabs component with the keyboard, both <kbd>Tab</kbd> and the arrow keys are not functioning as expected

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

Issue #980, opened 2022-03-09
[carbon-web-components]: migrate repository to monorepo


kennylam:
Migrate carbon-web-components repository into monorepo

This includes:
- re-target Jenkins jobs to Carbon for IBM.com repo
- update CDN paths
- archive `carbon-web-components` repo and update README to reflect change and point to new location

Issue #707, opened 2021-10-01
[tabs] Keyboard navigation is incomplete, broken in chrome - DDS Consulting


andy-blum:
###  Engineering info:
- HC engineer: 
- HC Jira Ticket: 
- DDS consulting engineer: 

----

## Detailed description

> Describe in detail the issue you're having.

Tabbed interfaces using `role="tablist"` and `role="tabitem"` carry an implicit interaction model that needs to be followed for WCAG compliance. Right now, that UI pattern is incomplete and broken in Chrome.

> Is this a feature request (new component, new icon), a bug, or a general
> issue?

:bug: Bug

> Is this issue related to a specific component?

Tabs

> What did you expect to happen? What happened instead? What would you like to
> see changed?

1. Missing keyboard interactions:
- While focus is within the tablist, the `Home` key should move focus to the **first** tab and activate it.
- While focus is within the tablist, the `End` key should move focus to the **last** tab and activate it.
- While focus is within the tablist, the `Tab` key should move focus to the active tabpanel element, and then to any focusable elements within that tabpanel before moving on to other parts of the page

2. Chrome's keyboard interaction model is completely broken at the moment. For some reason the links within each `<bx-tab>` are unable to receive/hold focus.

> What browser are you working in?

Tested in the following:
- Chrome 94.0.4606.61
- Firefox 92.0.1
- Safari 15.0 (16612.1.29.41.4, 16612)

> What version of Carbon Web Components are you using?

Latest at time of writing ([#01b2892](https://github.com/carbon-design-system/carbon-web-components/commit/01b289238cd9a8edceaa3ad1ee548cb6693a5da8))

> What offering/product do you work on? Any pressing ship or release dates we
> should be aware of?

carbon-for-ibm-dotcom, but we've already fixed this issue [downstream](https://github.com/carbon-design-system/carbon-for-ibm-dotcom/pull/7188) for now.

## Additional information

- [W3: Tabs example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/tabs/tabs-1/tabs.html#kbd_label)

RobertaJHahn:
@Praveen-sr Could you have Ram take a look to verify this bug and set the severity.  Thanks.

RobertaJHahn:
Praveen and Ram have started the testing.

andy-blum:
@annawen1 if `tabs-extended` isn't inheriting from this repo's `tabs` then this issue is only related in the sense that they should have similar solutions.

[...]

PR #992, opened 2022-03-21
fix(tabs): basic keyboard navigation

IgnacioBecerra:
### Related Ticket(s)
#970

### Description
This PR ensures that the Tabs and Content Switcher components are able to be work with tab navigation.

### Changelog

**New**

- added `tab-index` in the `tab` shadow dom

**Changed**

- triggering the content switching upon pressing the enter key

**Removed**

- remove unneeded `inFocus` property in the tab item component

### DIFF
--- a/src/components/content-switcher/content-switcher-item.ts
+++ b/src/components/content-switcher/content-switcher-item.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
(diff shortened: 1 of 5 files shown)


kennylam:
@IgnacioBecerra once #980 is complete we can also incorporate `home`/`end` interactions. See https://github.com/carbon-design-system/carbon-web-components/issues/707.

PR #1009, opened 2022-06-14
feat(link): add `icon-placement` option

emyarod:
### Related Ticket(s)

https://github.com/carbon-design-system/carbon-for-ibm-dotcom/issues/8836

### Description

This PR adds support for `icon-placement` in the link component to potentially deprecate the `link-with-icon` dotcom component

### DIFF
--- a/src/components/link/link-story.ts
+++ b/src/components/link/link-story.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2021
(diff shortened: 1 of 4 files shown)


emyarod:
@ariellalgilmore @RichKummer the knob should be on the `Paired with icon` story only now! also fixed the extra spacing when the icon is placed on the left side. and for the Percy changes, it looks like the snapshots are all brand new

RichKummer (approved):
LGTM! Thanks for fixing @emyarod !

PR #1036, opened 2022-10-01
fix(dropdown): :bug: fixing dropdown item disabled bug, resolves #1021

### LINKED ISSUE #1021: [bug]: `bx-dropdown-item` disabled property not working
### Description

When passing in the `disabled` property into the `bx-dropdown-item` component, I expected the item to be non-selectable, as an option.

However, as shown in the CodeSandbox example, this is not the case. A user is still able to select a `disabled` item.

### Component(s) impacted

`bx-dropdown-item`

related:
`bx-dropdown`

### Browser

Chrome

### Carbon Web Components version

1.21.0

### Severity

Severity 3 = The problem is visible or noticeable to users but does not impede the usability or functionality. Affects minor functionality, has a workaround.

### Application/website

IBM Quantum internal application 

### CodeSandbox example

https://codesandbox.io/s/spring-thunder-kgiidy?file=/index.html

### Steps to reproduce the issue (if applicable)

_No response_

### Release date (if applicable)

_No response_

### Code of Conduct

- [X] I agree to follow this project's [Code of Conduct](https://github.com/carbon-design-system/carbon-web-components/blob/main/.github/CODE_OF_CONDUCT.md)
- [X] I checked the [current issues](https://github.com/carbon-design-system/carbon-web-components/issues) for duplicate issues

brianferry:
### Related Ticket(s)

carbon-design-system/carbon-for-ibm-dotcom#9754 

### Description

Looking at this issue it appeared that the first issue was that there was a check for if the parent has a disabled attribute and then all of the child elements would be set to the parents disabled attribute (even if it was undefined).  The second issue I saw after this was related to being able to select a disabled dropdown item with the keyboard navigation. 

### Tested Browsers
- Firefox for Fedora - 105.0.1 (64-bit)
- Chromium - Version 103.0.5060.114 (Official Build) Fedora Project (64-bit)

### Changelog

**Changed**

- The first fix is  the line in `dropdown.ts` 

```js
   if (changedProperties.has('disabled')) {
```

to

```js
   if (!changedProperties.get('disabled')) {
```

so if it is undefined or set explicitly to false the child elements will not have their disabled attributes updated.

- Fix to being able to select a disabled bx-dropdown-item with the keyboard navigation inside of a bx-dropdown element.  Fixed by updating the function 

```js
   static get selectorItem() {
      return `${prefix}-dropdown-item`;
   }
```

```js
   static get selectorItemEnabled() {
      return `${prefix}-dropdown-item:not([disabled])`;
   }
```

and updating the references to this function in `dropdown.ts`.

### DIFF
--- a/src/components/dropdown/dropdown.ts
+++ b/src/components/dropdown/dropdown.ts
@@ -1,7 +1,7 @@
 /**
  * @license
  *
- * Copyright IBM Corp. 2019, 2021
+ * Copyright IBM Corp. 2019, 2022
  *
  * This source code is licensed under the Apache-2.0 license found in the
  * LICENSE file in the root directory of this source tree.
@@ -115,7 +115,9 @@ class BXDropdown extends ValidityMixin(HostListenerMixin(FormMixin(FocusMixin(Li
     if (this.shadowRoot!.contains(event.target as Node)) {
       this._handleUserInitiatedToggle();
     } else {
-      const item = (event.target as Element).closest((this.constructor as typeof BXDropdown).selectorItem) as BXDropdownItem;
+      const item = (event.target as Element).closest(
+        (this.constructor as typeof BXDropdown).selectorItemEnabled
+      ) as BXDropdownItem;
       if (this.contains(item)) {
         this._handleUserInitiatedSelectItem(item);
       }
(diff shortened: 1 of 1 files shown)


asudoh (approved):
LGTM 👍 from my perspective - I'm under an impression that the team is woking on moving this repo to another place, though.

